PR #2, opened 2015-10-08
Wrap missing headers

visr:
This is enough to be able to include common.jl without errors,
however the other files still miss types/constants.


visr:
For the GDAL Python SWIG wrapper this `setup.cfg` file is used:
```
# You can override default build options here

[build_ext]
#include_dirs = ../../port:../../gcore:../../alg:../../ogr/ 
#library_dirs = ../../.libs:../../ 
#libraries = gdal 
gdal_config=../../apps/gdal-config
```

(in addition to a few extensions)

yeesian:
This is great. I think the rest is mild leaky abstractions. With the new changes, the errors seem to be stemmed. It'll be good to try and wrap the remaining files mentioned in https://github.com/visr/GDAL.jl/commit/8540cc12814ef43b45fd30113bcbbfa5a5b11de7, and hope it ends there:

- https://github.com/OSGeo/gdal/blob/trunk/gdal/port/cpl_minixml.h
- https://github.com/OSGeo/gdal/blob/trunk/gdal/alg/gdalwarper.h
- https://github.com/OSGeo/gdal/blob/trunk/gdal/port/cpl_progress.h

visr:
Cool, now we're getting somewhere. I can't find anything on `FILE` or where it should reside, but your solution should work.

yeesian:
`FILE` is a generic C thing; not specific to GDAL

visr:
I see. Does this look good to you? Shall I merge?

src/clang/misc.jl
```diff
@@ -0,0 +1,6 @@
+typealias FILE Void # not sure if this works
+
+# workaround error in common.jl
+# https://github.com/OSGeo/gdal/blob/d7451d90134f6ca4dea8551b3a5d8591d8e4efad/gdal/port/cpl_port.h#L512-L534
+# this is wrapped as: const CPL_LSBPTR16 = x
+x = nothing
```
yeesian: This doesn't seem right.. Haha. The definitions are over at https://github.com/OSGeo/gdal/blob/d7451d90134f6ca4dea8551b3a5d8591d8e4efad/gdal/port/cpl_port.h#L442-L491. Perhaps comment them out, or hardcode the structs?

yeesian:
If you're happy, it should be good for merging. Squash+rebase maybe?

src/clang/misc.jl
visr: Hehe I wanted to keep the generated files without manual edits, so I used this. Though it's not very clean. Perhaps we can automatically catch these in `wrap_gdal.jl` later.

yeesian: Yeah, it's good to be clean, but not to be wrong though :wink:

Issue #4, opened 2015-11-21
Provide build script


yeesian:
for installing dependencies.

visr:
Is it mostly applying https://github.com/wkearn/RasterIO.jl/pull/11 to this repo, or would that not suffice?

yeesian:
Yup, applying what we did for RasterIO to this repo. This repo will then go into RasterIO's REQUIRE file.

PR #8, opened 2015-11-21
update readme for `C` submodule

yeesian:


visr:
Yes that's better. The C submodule should be doable to keep fully auto-generated. For the rest I'm not so sure yet, perhaps only as a bootstrap.

Ignore Travis, I just turned it on because I'm working on #4

PR #9, opened 2015-11-21
Use Conda as binary provider

### LINKED ISSUE #4: Provide build script

visr:
Closes #4.
@meggart, since this was basically copy paste of your work, you are author of the commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/deps/deps.jl
 *.jl.cov
 *.jl.*.cov
 *.jl.mem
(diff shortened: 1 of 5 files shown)


yeesian:
LGTM

visr:
I added another commit. If Pkg.build("GDAL") runs twice, it passes on Linux at least. However it finds a rather old GDAL:
```
INFO: Testing GDAL
WARNING: GDAL.jl is made for GDAL 2.0 and later
INFO: GDAL 1.9.2, released 2012/10/08
INFO: 113 GDAL drivers found
INFO: 53 OGR drivers found
INFO: GDAL tests passed
```

meggart:
I think this is as good as we can get right now, sooner or later we will have to find out why the double Pkg.build is necessary. Do you want to merge as is?

PR #10, opened 2015-11-24
update list of aliases

yeesian:
@joa-quim: this library will eventually be the common dependency for both RasterIO and OGR-bindings, so I propose moving over the changes from https://github.com/wkearn/RasterIO.jl/pull/21 to this repository. You okay with that?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,7 +3,8 @@ using Conda
 
 @BinDeps.setup
 
-libgdal = library_dependency("libgdal", aliases=["gdal", "gdal111"])
+libgdal = library_dependency("libgdal", aliases=["gdal","gdal111","gdal201",
+                                                 "gdal_w32","gdal_w64"])
 
 # install older gdal on windows
 @windows_only provides(Conda.Manager, "libgdal==1.11.2", libgdal)


joa-quim:
Yes, fine with me.

visr:
Hmm, odd, the latest GMT release comes with a `gdal_w64.dll` from the in-development GDAL 2.1.0:
```
julia> Pkg.test("GDAL")
INFO: Testing GDAL
INFO: GDAL 2.1.0dev, released 2015/99/99
INFO: 183 GDAL drivers found
INFO: 64 OGR drivers found
INFO: GDAL tests passed
```

No reason not to include it however.

joa-quim:
Yes, I normally use quite updated GDAL versions in GMT (it's me who build the Win GMT installers) 

BTW, I see that you did not suffer from

   http://gmt.soest.hawaii.edu/boards/1/topics/2829

visr:
Ha ok it's you that creates them. I did indeed not suffer from that issue. I'll probably have a look at GMT.jl sometime soon.

Merging this now.

meggart:
We might even think about using the GMT gdal dll files for Windows. Is there a way to download the gdal_w00.dll files directly, without using an installer?

visr:
There is, for instance the "Compiled binaries in a single .zip package" downloads from  http://www.gisinternals.com/. These are quite complete. Would be easiest if we could just use conda to get GDAL 2 though.

joa-quim:
The gisinternals zip has currently the gdal201.dll file. That is why I included it in the aliases list. But that zip file has a lot of files, many of them do not interest the gdal.dll directly, namely the exes and those relative to mapserver. It would be needed to carefully identify all the dependencies of the gdal.dll and those only. Doable, but not obvious.

Regarding the gdal version shipped  by GMT I could try to create for example a zip file with all the needed dependencies.

Issue #5, opened 2015-11-21
Provide docstrings


yeesian:
Given that we automatically check null for some of the wrapped functions, and provide some of our own types

yeesian:
After my experience writing documentation for RasterIO, we'll probably want to automate bits of this by scraping from gdal's documentation before modifying them into docstrings.

visr:
I'm working on automating this, after working on it a bit more I'll open a PR.

Right now I've taken the Doxyfile from the GDAL SVN, set `GENERATE_XML = YES`, then combined all resulting files in one 20MB XML. With this I can lookup the function name in the XML and add the part of the documentation I want during the rewriting. For now simply the brief description:

```julia
functionnode(et, fname) = et["/doxygen/compounddef/sectiondef/memberdef[name='$fname']"]
briefdesc = fnode["briefdescription/para/text()"]

"Find out if we have update permission for this band. "
function GDALGetRasterAccess(arg1::GDALRasterBandH)
    ccall((:GDALGetRasterAccess,libgdal),GDALAccess,(GDALRasterBandH,),arg1)
end
```

It needs more work to nicely display more extensive documentation in markdown, like this: 
http://gdal.org/gdal_8h.html#a194a45dd5319e8f0631d74a9faa518da

yeesian:
\o/ That will be great!

yeesian:
closed by #11

PR #11, opened 2015-12-25
auto-generated documentation

visr:
Todo:
- [x] Convert to Markdown syntax
- [x] Settle on a good format that also handles missing fields
- [x] Add these docs to the main module's generated functions as well

It coud be confusing that the input parameters are presented as C types. We could decide to replace it by the Julia conversion, but often these are unhelpful `arg1`, `arg2` as in:
```julia
"""
    GDALTermProgress(CPL_UNUSED double dfComplete, CPL_UNUSED const char *pszMessage, void *pProgressArg)

Simple progress report to terminal.

Returns int
"""
function GDALTermProgress(arg1::Cdouble,arg2::Ptr{UInt8},arg3::Ptr{Void})
    ccall((:GDALTermProgress,libgdal),Cint,(Cdouble,Ptr{UInt8},Ptr{Void}),arg1,arg2,arg3)
end
```

Part of #5

### DIFF
--- a/gen/wrap_gdal.jl
+++ b/gen/wrap_gdal.jl
@@ -1,4 +1,9 @@
 using Clang
+using LibExpat
+
+workdir = dirname(@__FILE__)
+srcdir = joinpath(workdir, "../src")
+include(joinpath(workdir, "doc.jl"))
 
 gdalpath = "/usr/local/include"
 includedirs = [gdalpath]
@@ -39,10 +44,17 @@ function minimal_rewriter(ex::Expr)
     # if it is a function
     if ex.head == :function
         funcname = ex.args[1].args[1]
+        @show funcname
         # skip predefined functions
         if funcname in skip_func
             return ""
         end
+        # add docstring
+        fnode = functionnode(et, string(funcname))
+        docstr = build_docstring(fnode)
+        # turn it into a tuple, special cased in visr/Clang.jl
+        # done because a block Expr gets a begin/end block when printing
+        ex = (docstr, ex)
     end
     ex
 end
@@ -51,7 +63,7 @@ minimal_rewriter(s::AbstractString) = s # keep comments
 minimal_rewriter(A::Vector) = [minimal_rewriter(a) for a in A]
 
 context=wrap_c.init(headers=headerpaths,
-                    output_dir="src/C",
+                    output_dir=joinpath(srcdir, "C"),
                     common_file="common.jl",
                     clang_includes=includedirs,
                     header_wrapped=header_wrapped,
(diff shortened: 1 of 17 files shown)


yeesian:
this is quite a christmas gift (: :christmas_tree: :tada:

visr:
And with this, Santa learned how to reproject his map to the familiar polar projection, and Christmas was saved.

gen/wrap_gdal.jl
```diff
@@ -23,6 +24,28 @@ skip_expr = [:(const CPL_LSBPTR16 = x),
 
 skip_func = [:CPLErrorV] # problem with va_list ihnorton/Clang.jl#17
 
+using LibExpat
+# This is a testing script to see if we can auto generate markdown docstrings
+# for the GDALDatasetCreateLayer function based on the XML output from
+# GDAL's Doxygen documentation.
+# This was combined in one XML by running combine.xslt
+xmlfile = "gen/all.xml"
```
yeesian: how big is the resulting file? Should we check it into this repo?

yeesian:
> It coud be confusing that the input parameters are presented as C types.

Given that this is already fairly low-level (and unexported), I'm in favour of just keeping the raw C description for the `C/` folder. People who use functions from there are expected to be familiar with the C API. For functions where we automatically check null, that has to be mentioned in the docstring as well.

visr:
I just pushed an updated version. The generated docstrings don't contain all information like on GDAL.org, but have all the basics more or less the way you suggested.

If you want to play around with it yourself, I added some documentation to `gen/README.md`. If you want to see the XML with all the documentation, I uploaded it [here](https://s3.eu-central-1.amazonaws.com/visr/julia/gdal/gdal-doxygen-xml.zip). It's rather large to commit it as XML to the repo.

> For functions where we automatically check null, that has to be mentioned in the docstring as well.

Wouldn't it be better to tell something about GDAL.jl's error handling in one place? I'm no longer sure if we should keep the `checknull` function around anyway, because we now have the custom error handling that throws on any call from GDAL, also where GDAL would normally continue. This means however that opening non existent files will throw, but getting nonexistent drivers will only return a NULL:

https://github.com/visr/GDAL.jl/blob/f70a9aeb204e01a459395dc3b6339b2051490e07/test/runtests.jl#L12-L15

yeesian:
> I'm no longer sure if we should keep the checknull function around anyway, because we now have the custom error handling that throws on any call from GDAL, also where GDAL would normally continue. This means however that opening non existent files will throw, but getting nonexistent drivers will only return a NULL

Oh okay I thought the final decision from https://github.com/visr/GDAL.jl/pull/3 was to keep `checknull`, since it wasn't removed before merging. I'm in favour of removing it from the base C functions that we provide too. This should also be dealt with (or documented) at some point; you don't have to worry about it in this PR though.

yeesian:
LGTM. Closes #5

PR #12, opened 2016-03-11
regenerate branch

### LINKED ISSUE #7: Wrap pointers into julian types
We should define e.g.

```julia
immutable GDALDriver
    ptr::GDALDriverH
end
```

for each of the typealiases in https://github.com/visr/GDAL.jl/blob/22b70b7b74ac75d645c5361e779a38474df860a3/gen/rewriter.jl#L29-L54

And the methods in `src/gdal_*` and `src/ogr_*` should be defined on these types, rather than on `Ptr{Void}`.

Should this be done in the rewriter?

visr:
This became a bit of a mix of different things that is hard to untangle now, I'll try to separate things better in the future.

Below are the most significant parts.

##### Regenerate with GDAL trunk
Because I would like to wrap the new `gdal_utils.h` [RFC 59.1](https://trac.osgeo.org/gdal/wiki/rfc59.1_utilities_as_a_library), and would rather not wait until the GDAL 2.1 release. This only re-wraps the original header files, wrapping the utils will be done later in a separate PR.

##### Adding tests
These are only some tests based on the API tutorials. Part of #6.

##### Wrap pointers into julian types
Closes #7. The approach here is as discussed in that issue. So:
```julia
abstract GDALDatasetH <: GDALMajorObjectH

function getrasterband{T <: GDALDatasetH}(arg1::Ptr{T},arg2::Integer)
    checknull(ccall((:GDALGetRasterBand,libgdal),Ptr{GDALRasterBandH},(Ptr{GDALDatasetH},Cint),arg1,arg2))
end
```


visr:
Tests failing because the Travis tests use GDAL 1.9 on Linux right now, which doesn't have `GDALOpenEx`. But I wasn't planning to support anything pre 2.0, so this is a matter of fixing the Travis setup.

yeesian:
Thanks for this! I'll give it a spin over the weekend. Anything else left pending for this PR?

yeesian:
@meggart @joa-quim

gen/rewriter.jl
```diff
@@ -11,6 +14,11 @@ headerfiles = [
     "ogr_srs_api.jl"
 ]
 
+custom_rename = Dict{ASCIIString, ASCIIString}(
+    "gdalgetdriverbyname" => "gdalgetdriverbyname",
+    "ogrgetdriverbyname" => "ogrgetdriverbyname"
+)
+
```
yeesian: what is the purpose of `custom_rename`? I've only seen it used in `newfuncname` as

```julia
if name in keys(custom_rename)
    name = custom_rename[name]
else
    ...
```

gen/wrap_gdal.jl
```diff
@@ -5,7 +5,7 @@ workdir = dirname(@__FILE__)
 srcdir = joinpath(workdir, "../src")
 include(joinpath(workdir, "doc.jl"))
 
-gdalpath = "/usr/local/include"
+gdalpath = "/home/martijn/bin/gdal/include"
```
yeesian: Maybe leave a note in the `GDAL.jl generation documentation` for developers to update this `gdalpath` variable to their respective locations?

src/C/gdal_alg.jl
```diff
@@ -869,20 +947,31 @@ Burn geometries into raster.
...
+												"BURN_VALUE_FROM": 
+											
+											
+												May be set to "Z" to use the Z values of the geometries. dfBurnValue is added to this before burning. Defaults to GDALBurnValueSrc.GBV_UserBurnValue in which case just the dfBurnValue is burned. This is implemented only for points and lines for now. The M value may be supported in the future. 
+											
+											
+												"MERGE_ALG": 
+											
+											
+												May be REPLACE (the default) or ADD. REPLACE results in overwriting of value, while ADD adds the new value to the existing raster, suitable for heatmaps for instance.
```
yeesian: the whitespacing in the autogenerated docstrings seems a little off here

src/GDAL.jl
```diff
@@ -30,7 +30,8 @@ module C
 end
 
 include("C/misc.jl")
-include("C/common.jl")
+include("types.jl")
+include("common.jl")
```
yeesian: Should the aliases in `common.jl` be considered as part of the (low-level) C interface? My understanding's that

- they also contain a lot of typealiases like `ANONYMOUS_15`
- we might want to reserve `src/common.jl` for our own keywords

yeesian:
I ran into the following blocker:

```julia
  | | |_| | | | (_| |  |  Version 0.4.2 (2015-12-06 21:47 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> import GDAL

help?> GDAL.getdriver
  OGR_DS_GetDriver(OGRDataSourceH) -> OGRSFDriverH

  Returns the driver that the dataset was opened with.
  [...]

julia> GDAL.getdriver(3)
Ptr{GDAL.OGRSFDriverH} @0x00007fd50d5574e0

julia> GDAL.getdrivershortname(GDAL.getdriver(3))
ERROR: MethodError: `getdrivershortname` has no method matching getdrivershortname(::Ptr{GDAL.OGRSFDriverH})

help?> GDAL.getdrivershortname
  GDALGetDriverShortName(GDALDriverH hDriver) -> const char *

  [...]
```

Correct me if I'm wrong, but I think the signature for GDAL.getdrivershortname should take in the pointer to `GDAL.OGRSFDriverH`?

yeesian:
We need to do a comprehensive job of mapping out the type hierarchy, adding the other objects, e.g.
```julia
abstract OGRSFDriverH <: GDALDriverH
```
to `types.jl`, and rerunning `rewriter.jl`. Otherwise we'll run into issues like
```julia
  | | |_| | | | (_| |  |  Version 0.4.2 (2015-12-06 21:47 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> import GDAL

help?> GDAL.getdriver
  OGR_DS_GetDriver(OGRDataSourceH) -> OGRSFDriverH
  [...]

julia> GDAL.getdriver(3)
Ptr{GDAL.OGRSFDriverH} @0x00007f91c63cc600

julia> GDAL.getdrivershortname(ans)
ERROR: MethodError: `getdrivershortname` has no method matching getdrivershortname(::Ptr{GDAL.OGRSFDriverH})

help?> GDAL.getdrivershortname
  GDALGetDriverShortName(GDALDriverH hDriver) -> const char *
```

gen/README.md
```diff
@@ -4,14 +4,16 @@ GDAL.jl generation documentation
...
 this fork of Clang.jl is used: https://github.com/visr/Clang.jl (branch gdal)
+For the generation process `julia 0.5-dev` based on `LLVM 3.7.1` and with
+`BUILD_LLVM_CLANG=1` in `Make.user` is used.
 
 ### Make sure you have the GDAL header files
-The `wrap_gdal.jl` script looks in `/usr/local/include` by default.
+The `wrap_gdal.jl` script finds them in the directory set in the `gdalpath` variable.
 
 ### Prepare the GDAL Doxygen XML
 1. In the GDAL SVN, run Doxygen with `GENERATE_XML = YES`
```
yeesian: Navigate to the folder containing GDAL SVN, and run Doxygen with `GENERATE_XML = YES` with the command
```
$ doxygen Doxyfile GENERATE_XML = YES
```

gen/combine_gdal_doxygen_xml.py
```diff
@@ -0,0 +1,37 @@
...
+# place in directory with all xmls created by doxygen
+# this is by default gdal/xml/
+
+# done previously by running combine.xslt
+# however the index.xml doesn't seem up to date
+# with the contents of gdal/xml/
+
+# after this script, doxygen.xml will be around 80 MB
+# I manually deleted everything not under
+# sectiondef, briefdescription, and detaileddescription
```
yeesian: How do I do that?

gen/README.md
```diff
@@ -4,14 +4,16 @@ GDAL.jl generation documentation
...
-2. In the `xml` output dir, run `combine.xslt` to combine all documentation in one XML
-3. Copy this file to `GDAL.jl/gen/all.xml`
+2. Follow the instructions in `combine_gdal_doxygen_xml.py`
+3. Copy the result to `GDAL.jl/gen/doxygen.xml`
```
yeesian: I've managed to follow all the steps, apart from manually modifying the xml file, and am getting errors like
```bash
$ julia gen/rewriter.jl
ERROR: LoadError: LoadError: "ErrorException(\"Error parsing document : 0\"), mismatched tag, 128156, 3, 8208979"
 in error at /Applications/Julia-0.4.2.app/Contents/Resources/julia/lib/julia/sys.dylib
 in xp_parse at /Users/yeesian/.julia/v0.4/LibExpat/src/LibExpat.jl:289
```
Is it okay host a copy of this generated file somewhere available in the meantime?

gen/rewriter.jl
```diff
@@ -85,48 +114,107 @@ function rewriter(ex::Expr)
     # turn it into a tuple, special cased in visr/Clang.jl
     # done because a block Expr gets a begin/end block when printing
 
+    partypes = OrderedDict{Symbol, Symbol}()
```
yeesian: Maybe leave a note on what `partypes`, `letters`, and `letter_index` are about?

yeesian:
I started a new package over at https://github.com/yeesian/GDALUtils.jl (based off @wkearn's RasterIO package) just to check out the API from this package, and am enjoying it quite abit.

In particular, some of the functions in GDAL.jl might be over-eager in their evaluation of gdal's output, e.g. [`getfilelist`](https://github.com/visr/GDAL.jl/blob/01b78a8576e32a8fb10451ffec4bc552f72d7270/src/gdal_h.jl#L698) which evaluates an array of strings as a single `bytestring`. It's good that we have C-fallbacks, allowing me to work around it: https://github.com/yeesian/GDALUtils.jl/blob/master/src/metadata.jl#L55-L56.

test/tutorial_vector.jl
```diff
@@ -0,0 +1,78 @@
...
+
+# TODO: provide a convenient way to construct nOpenFlags
+# nOpenFlags: a combination of GDAL_OF_ flags that may be combined through logical or operator
+# Driver kind: GDAL_OF_RASTER for raster drivers, GDAL_OF_VECTOR for vector drivers. If none of the value is specified, both kinds are implied.
+# Access mode: GDAL_OF_READONLY (exclusive)or GDAL_OF_UPDATE.
+# Shared mode: GDAL_OF_SHARED. If set, it allows the sharing of GDALDataset handles for a dataset with other callers that have set GDAL_OF_SHARED. In particular, GDALOpenEx() will first consult its list of currently open and shared GDALDataset's, and if the GetDescription() name for one exactly matches the pszFilename passed to GDALOpenEx() it will be referenced and returned, if GDALOpenEx() is called from the same thread.
+# Verbose error: GDAL_OF_VERBOSE_ERROR. If set, a failed attempt to open the file will lead to an error message to be reported.
+
+# file from GDAL SVN
+dataset = GDAL.openex("data/point.geojson", UInt32(GDAL.GDAL_OF_VECTOR), nil, nil, nil)
```
yeesian: We should relax the function definition for `GDAL.C.GDALOpenEx` i think

gen/rewriter.jl
visr: Yes it appears to be doing nothing, but because of the if/else construct, it prevents automatic renaming and overwriting method definitions for these two. I'll add a comment that clarifies this.

gen/combine_gdal_doxygen_xml.py
visr: I used XMLSpy for this, but this is not free software. I'll add some code to the end of the python script that does the same. Note that this step is not essential, but it does speed up the wrapping since it takes less time to lookup documentation.

gen/README.md
visr: Here is the version I am using:
https://s3.eu-central-1.amazonaws.com/visr/julia/gdal/doxygen.xml
I'll add this link to the instructions as well. But cool that you are trying it yourself as well. From the error it seems the XML you are using is not well-formed (mismatched tag), did you check this?

yeesian: Yup, I realized and did check. It was when I thought it's probably better that we find ways of making it easier for subsequent developers to go through the process of regenerating the documentation. Thanks again for the instructions! I wouldn't have made it that far without them.

test/tutorial_vector.jl
visr: What are you proposing? Compared to master it is already relaxed in this branch:
https://github.com/visr/GDAL.jl/blob/d76caa2263fbd64560bb1057a11698e3b925b9ee/src/C/gdal.jl#L313-L315

yeesian: Might relaxing `UInt32` to `Integer` work for `GDAL.openex`?

visr: Agreed, done.

src/C/gdal_alg.jl
visr: Not sure where the leading whitespace came from, but now it's gone after using the updated python script to generate the XML.

meggart:
I think we should merge as soon as possible and then start porting RasterIO and (and OGR-related stuff?) to the new interface. I think this will be a good test case and we can go back whenever problems appear.

src/GDAL.jl
visr: I just added commit 34e16701c0c89a717a8c14bd73476e1e57ad4102.
This removes all `typealias ANONYMOUS_*` from `src/common.jl`.

Currently the main module is not dependent on the C submodule, and since quite a few of the constants are needed in normal use (like `GDAL.GDAL_OF_VECTOR`), I'm not sure whether we should put all this in the C submodule.

Perhaps I don't fully understand and you can clarify. Are you talking only about the typealiases in `src/common.jl`, or about the whole file. We can also do a better job rewriting `src/common.jl`. `src/types.jl` is basically already a manual rewrite of parts of common.

visr:
Thanks for the review, @yeesian.

>In particular, some of the functions in GDAL.jl might be over-eager in their evaluation of gdal's output, e.g. getfilelist which evaluates an array of strings as a single bytestring. It's good that we have C-fallbacks, allowing me to work around it.

You are right that the current behavior is wrong. Do you think we should just return `Ptr{Cstring}` in the main module, or process it to `Vector{UTF8String}`?

@meggart agree that we can merge this soon. The `bytestring` fix might still be good to include, but other than that I'm finished with this PR (unless I missed something in the commits that I have added, @yeesian).

src/GDAL.jl
yeesian: I'm talking about the whole file.

I agree that quite a few of the constants are needed in normal use, but I can't help but feel they might be better off as symbols or immutables instead (makes them a lot more interpretable when rendered in the REPL, rather than becoming obscure constants). Hence my impression that the constants should be in `src/C/common.jl`, and the corresponding symbols/mappings (e.g. https://github.com/yeesian/GDALUtils.jl/blob/master/src/metadata.jl#L24) in `src/common.jl`.

yeesian:
Please feel free to merge whenever you feel ready, it has addressed most/all of my concerns.

> Do you think we should just return `Ptr{Cstring}` in the main module, or process it to `Vector{UTF8String}`?

I think we should process it to `Vector{UTF8String}`. Luckily there isn't a lot of them, so special-casing them, and manually identifying them on a case-by-case basis should be good enough to sift out the important ones.

It shouldn't be a blocker for this PR though. We can file a separate issue for it after the PR gets merged.

yeesian:
> I think we should merge as soon as possible and then start porting RasterIO and (and OGR-related stuff?) to the new interface. I think this will be a good test case and we can go back whenever problems appear.

That work has already partially begun at [GDALUtils](https://github.com/yeesian/GDALUtils.jl/) (suggestively named for it to rename experimental in testing out ideas/syntax). Feel free to copy ideas/code/etc from there over to this repository.

PR #17, opened 2016-03-17
disambiguate create() functions

### LINKED ISSUE #15: Disambiguate create() functions
Relevant lines: https://github.com/visr/GDAL.jl/blob/master/src/ogr_api.jl#L2348-L236 and https://github.com/visr/GDAL.jl/blob/master/src/ogr_api.jl#L2764-L2766

To be consistent with `fld_create`, and `gfld_create` (`create` is too general a name, and might have conflicts with creating other GDAL objects). I think we should add `ogr_fd_create` and `ogr_f_create` to the list at https://github.com/visr/GDAL.jl/blob/master/gen/rewriter.jl#L24-L26, and rerun the rewriter code.

visr:
closes #15

These functions did not have clashing signatures,
but `create` is not a very helpful name.

Now the only `create` function left is `GDALCreate`.

### DIFF
--- a/gen/rewriter.jl
+++ b/gen/rewriter.jl
@@ -25,6 +25,15 @@ custom_rename = Dict{ASCIIString, ASCIIString}(
     "ogr_fld_create" => "fld_create",
     "ogr_gfld_create" => "gfld_create",
 
+    # these don't directly clash, but otherwise all are renamed to `create`
+    # this makes it easier to disambiguate them
+    "ogr_fd_create" => "fd_create",
+    "ogr_f_create" => "f_create",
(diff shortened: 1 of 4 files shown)


yeesian:
LGTM. Just a question: what do you think about methods like
- `creategeometry`
- `creategeomfield`

(since they came up in https://github.com/yeesian/GDALUtils.jl/issues/2)

visr:
Hmm yeah these seems like some inconsistencies in the GDAL naming. Both have at least been renamed to something a little bit more Julian. I currently think it's not for GDAL.jl to try to make it all consistent. The big leap in making a Julian package will not be made in this package anyway.

PR #22, opened 2016-05-24
update to GDAL 2.1.0 release

visr:
From now on I'd like to stick to official releases if possible.
The usage of NTuple instead of immutable composites in the new Clang.jl cleans up `common.jl` a bit.


yeesian:
LGTM. It works for me too. Feel free to merge whenever you're ready.

PR #23, opened 2016-06-02
use conda-forge's gdal recipe

visr:
This conda channel has a GDAL recipe that works well on all platforms, is actively maintained, and has GDAL 2.1.

With the validate function in `build.jl` we won't accept GDAL 1.x anymore. This was around on Travis, causing builds to fail.

With this all is green, except Windows 32 bits, I'll open a separate issue about that.

Possible downside of using conda-forge gdal instead of default channel libgdal is that it's 100MB heavier: 

https://travis-ci.org/visr/GDAL.jl/jobs/132722834#L123-L139
https://travis-ci.org/visr/GDAL.jl/jobs/132740954#L123-L146

Mainly due to pulling in MKL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ notifications:
 # uncomment the following lines to override the default test script
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia -e 'Pkg.clone(pwd()); Pkg.build("GDAL"); Pkg.build("GDAL"); Pkg.test("GDAL"; coverage=true)'
+  - julia -e 'Pkg.clone(pwd()); Pkg.build("GDAL"); Pkg.test("GDAL"; coverage=true)'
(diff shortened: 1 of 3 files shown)


visr:
Most of this is superseded by #25 or included in #26. The Linux build is still using `Conda.jl` however, from the default channel. This is GDAL 2.0.0. Should we still add `Conda.add_channel("conda-forge")` to get 2.1 on Linux like we do on Windows? We could even make it the minimum required version, since we can provide it anyhow. Or do you want to move away from `Conda.jl` on unix as well @yeesian?

yeesian:
> Should we still add Conda.add_channel("conda-forge") to get 2.1 on Linux like we do on Windows?

Personally, I won't mind using Conda to get 2.1 on Linux. @tkelman?

tkelman:
Yeah, Conda-forge's toolchain is more likely to be compatible with Julia on Linux than on Windows.

PR #38, opened 2016-08-09
wrap gdal_utils.h

### LINKED ISSUE #37: Provide support for gdal utilities
Available for GDAL 2.1. Relevant header file: http://www.gdal.org/gdal__utils_8h.html

visr:
This provides an API that resembles the GDAL command line tools.

Didn't really try this out yet, but putting it out there.
Fixes #37.


yeesian:
:+1: I feel uncomfortable about the auto-generated (opaque) options types, and think we should merge only if we can demonstrate that it works. Thanks for this though!

visr:
I added some tests. The option types seem to work fine. Any other comments?

yeesian:
LGTM.

test/runtests.jl
```diff
@@ -26,7 +26,11 @@ cd(dirname(@__FILE__)) do
     include("tutorial_raster.jl")
     include("tutorial_vector.jl")
     include("tutorial_vrt.jl")
-    include("gdal_utils.jl")
+    # several conversions for ccall are only added in v0.5
```
yeesian: out of curiosity: what are the ccall conversions that were only added in v0.5?

test/gdal_utils.jl
```diff
@@ -0,0 +1,162 @@
...
+
+
+# GDALWarp
+
+# cannot reproject file
+# ERROR: LoadError: LoadError: GDALError (Failure, code 6):
+#        Unable to load PROJ.4 library (libproj.so), creation of OGRCoordinateTransformation failed.
+# when using
+# options = GDAL.warpappoptionsnew(["-t_srs","EPSG:4326"], C_NULL)
+# should we require PROJ.4 or not?
```
yeesian: our options here might be to (i) [statically link PROJ.4](https://lists.osgeo.org/pipermail/gdal-dev/2004-October/004332.html), or (ii) [set the environment path variable](http://gis.stackexchange.com/a/157274) (if the dll exists).

Possibly relevant: https://github.com/ContinuumIO/anaconda-issues/issues/263

test/runtests.jl
visr: If I remove the condition the first error I get is
```MethodError: `convert` has no method matching convert(::Type{Ptr{GDAL.GDALDatasetH}}, ::Array{Ptr{GDAL.GDALDatasetH},1})```
from https://github.com/visr/GDAL.jl/blob/453ba74b8b8e4f6292a80c4f1c4128f1e7f4a280/test/gdal_utils.jl#L61

Which is a bit odd since if I test on v0.4
```
julia> abstract Foo
julia> Base.cconvert(Ptr{Foo}, [Ptr{Foo}(C_NULL)])
1-element Array{Ptr{Foo},1}:
 Ptr{Foo} @0x0000000000000000
```
it seems to work fine. Since it was working on v0.5 I didn't want to go to the bottom of it.

But the function does seem to expect a list of input datasets.
https://github.com/visr/GDAL.jl/blob/453ba74b8b8e4f6292a80c4f1c4128f1e7f4a280/src/gdal_utils.jl#L200-L201

test/gdal_utils.jl
visr: Great, thanks for the links. I got the error on a minimal GDAL that I built myself. Would be good if we can ensure at the `Pkg.build` step that everybody has PROJ.4.

PR #41, opened 2017-08-29
regenerate with latest Julia/GDAL/Clang.jl

### LINKED ISSUE #36: Should we wrap cpl_conv?
relevant header file: http://www.gdal.org/cpl__conv_8h.html

I noticed it from https://trac.osgeo.org/gdal/wiki/ConfigOptions.

visr:
This includes many different things, the most important being:
- generate on Julia v0.6.0 (specifically `visr/julia-0.6+` for a print fix that's on master)
- the generated code works on v0.6.0 (it dit previously but due to manual fixes after generating)
- based on GDAL 2.2.1
- generated with `visr/Clang.jl`, which is now based on Clang.jl v0.3.0
- wrap `cpl_conv` & `cpl_virtualmem` & `cpl_vsi` (fixes #36, others were needed as well to fix errors)
- automatically check the completeness of `src/types.jl`, and use its supertypes
- remove handmade list of `Ptr{Void}`, generate it from `common.jl`
- use `Ptr{Void}` as input types, `Ref` in function signature, see [discourse](https://discourse.julialang.org/t/converting-ptr-of-subtype-to-ref-of-supertype-in-ccall/5620/) for discussion

I realize this is a too much at once for proper review, but I hope this will make the generation process more robust and easier to follow for the future.


visr:
Worth noting that this also drops julia v0.5 support, if that's a problem for anyone let me know.
@yeesian, perhaps you have time for a quick look?

yeesian:
Yeah sure I might take some time, but the changes here looks good, and I'd like to go through them too. Do you have any hard deadlines I should get it done by? Otherwise I'll do so this weekend.

visr:
No deadline no, take your time.

src/C/common.jl
```diff
@@ -20,29 +21,21 @@ const CPLE_AWSSignatureDoesNotMatch = 16
 
 # const CPLAssert = expr
 
-# begin enum CPLErr
-const CPLErr = UInt32
+# begin enum ANONYMOUS_1
+const ANONYMOUS_1 = UInt32
```
yeesian: Is it me, or do the generated enums look off? `CPLErr` should still be `UInt32`, rather than being rewritten as `ANONYMOUS_$i`. It's the same with the other enums as well.

src/C/common.jl
```diff
@@ -55,35 +48,33 @@ const CXT_Comment = (UInt32)(3)
...
 
-type CPLXMLNode
+mutable struct CPLXMLNode
     eType::CPLXMLNodeType
     pszValue::Cstring
-    psNext::Ptr{CPLXMLNode}
-    psChild::Ptr{CPLXMLNode}
+    psNext::Ptr{Void}
+    psChild::Ptr{Void}
 end
```
yeesian: I'm not sure about recursive types like these, since I haven't yet used and/or tested them

visr: Yeah me neither, I'll try them out / add a test before merging.

src/C/common.jl
visr: Yeah they are off. I added this code to deal with it when rewriting: https://github.com/visr/GDAL.jl/blob/d9cabd4295819b924bf9a753d0b2021880fba4eb/gen/rewriter.jl#L193-L199

But that's more of a patch to fix something that seems to be going wrong in Clang.jl. I realize however that this now makes the C submodule hardly usable, so let's try to see if we can fix this in Clang.jl before merging.

src/C/common.jl
visr: In the latest wrapping it was no longer self referential, but I added a test in 57dac89c11fc16ccb2bab7d3aa9ddcf307a97a92

src/C/common.jl
visr: `ANONYMOUS_$i` no longer occurs in our wrapping, due to a number of fixes in `visr/Clang.jl`, that still need to be upstreamed.
Enums are now wrapped as enums.

visr:
Took a while before I could get to it, but the mentioned issues are now resolved. @yeesian, mind taking another look?

src/C/gdalwarper.jl
```diff
@@ -16,15 +16,15 @@
...
+function GDALWarpNoDataMasker(pMaskFuncArg, nBandCount::Cint, eType::GDALDataType, nXOff::Cint, nYOff::Cint, nXSize::Cint, nYSize::Cint, papabyImageData, bMaskIsFloat::Cint, pValidityMask, pbOutAllValid)
+    ccall((:GDALWarpNoDataMasker, libgdal), CPLErr, (Ptr{Void}, Cint, GDALDataType, Cint, Cint, Cint, Cint, Ptr{Ptr{GByte}}, Cint, Ptr{Void}, Ptr{Cint}), pMaskFuncArg, nBandCount, eType, nXOff, nYOff, nXSize, nYSize, papabyImageData, bMaskIsFloat, pValidityMask, pbOutAllValid)
 end
 
 
 """
     GDALWarpDstAlphaMasker(void * pMaskFuncArg,
                            int nBandCount,
-                           GDALDataType eType,
+                           GDALDataType,
```
yeesian: seems to have dropped some argument names?

src/C/gdalwarper.jl
```diff
@@ -33,15 +33,15 @@ end
...
+    ccall((:GDALWarpDstAlphaMasker, libgdal), CPLErr, (Ptr{Void}, Cint, GDALDataType, Cint, Cint, Cint, Cint, Ptr{Ptr{GByte}}, Cint, Ptr{Void}), pMaskFuncArg, nBandCount, eType, nXOff, nYOff, nXSize, nYSize, arg1, bMaskIsFloat, pValidityMask)
 end
 
 
 """
     GDALWarpSrcAlphaMasker(void * pMaskFuncArg,
-                           int nBandCount,
-                           GDALDataType eType,
+                           int,
+                           GDALDataType,
```
yeesian: seems to have dropped some argument names?

yeesian: not important though, since it's just the docstring

yeesian:
is the most salient change the move to enums?

yeesian:
it also drops support for `julia0.5` right? Should we reflect that in travis and appveyor too?

visr:
Great, thanks! I'm surprised by the docstring regression, will see if I can figure it out.

> is the most salient change the move to enums? Are there other changes I should be aware of?
Nope, that's the one.

> it also drops support for julia0.5 right? Should we reflect that in travis and appveyor too?
You're right, just noticed it, will add that before merging.

src/C/common.jl
```diff
@@ -636,24 +412,22 @@ const OGRGeometryH = Ptr{Void}
...
+mutable struct _CPLXMLNode
 end
 
 const OGRFieldDefnH = Ptr{Void}
 const OGRFeatureDefnH = Ptr{Void}
 const OGRFeatureH = Ptr{Void}
 const OGRStyleTableH = Ptr{Void}
 
-type OGRGeomFieldDefnHS
+mutable struct OGRGeomFieldDefnH
```
yeesian: how are these meant to be used?

```julia
julia> GDAL.getnameref(GDAL.OGRGeomFieldDefnH())
ERROR: ccall: the type of argument 1 doesn't correspond to a C type
Stacktrace:
 [1] getnameref(::GDAL.OGRGeomFieldDefnH) at /Users/yeesian/.julia/v0.6/GDAL/src/ogr_api.jl:2506

julia> GDAL.gettype(GDAL.OGRGeomFieldDefnH())
ERROR: ccall: the type of argument 1 doesn't correspond to a C type
Stacktrace:
 [1] gettype(::GDAL.OGRGeomFieldDefnH) at /Users/yeesian/.julia/v0.6/GDAL/src/ogr_api.jl:2522
```

visr: Good point, don't know. I assumed that it would work since Clang.jl wraps them like this now, but perhaps I should manually force it back in `types.jl`, the old way.

I believe it is wrapped differently since in C it is `typedef struct OGRGeomFieldDefnHS *` as opposed to just `typedef void *`.

yeesian: It was something that surfaced in ArchGDAL as I was updating it. I think there's a `OGRGeomFieldDefnHS` and `OGRGeomFieldDefnH` is a pointer to it.

src/C/gdalwarper.jl
visr: I see this information is no longer in the doxygen generated XML that I use to retrieve this information:
```xml
        <param>
          <type>int</type>
          <declname>nBandCount</declname>
        </param>
        <param>
          <type><ref refid="gdal_8h_1a22e22ce0a55036a96f652765793fb7a4" kindref="member">GDALDataType</ref></type>
        </param>
```

gdalwarper.cpp:
```cpp
GDALWarpDstAlphaMasker( void *pMaskFuncArg, int nBandCount,
                        CPL_UNUSED GDALDataType /* eType */,
```

gdalwarper.h
```cpp
GDALWarpDstAlphaMasker( void *pMaskFuncArg, int nBandCount, GDALDataType eType,
                        int nXOff, int nYOff, int nXSize, int nYSize,
```

I guess this is left out by doxygen because of the `/* eType */` in the cpp file. Either that is new, or the current doxygen version behaves differently.

Perhaps something to pick up in the future.

yeesian: yeah sure

src/C/common.jl
visr: Could you check my edits in 62b243dbde35916de718738d6392b8fee89744b1?
Otherwise I believe this PR is ready.

yeesian: I can confirm that all the tests are passing locally on https://github.com/yeesian/ArchGDAL.jl/pull/20 based on the changes here now

yeesian: I should mention I'm very happy with the consistency in the pointer types now, and with the change to use enums

visr: I'm very happy with you pointing out any inconsistencies :)

Issue #42, opened 2017-08-30
BinDeps.PackageManager failed to satisfy dependency libgdal on 32bit Windows and Linux


visr:
Ever since 8087fea74eca384be44fc1b537a8aa570fc341f6 AppVeyor 32 bit builds have been failing with:
```
LoadError: Provider BinDeps.PackageManager failed to satisfy dependency libgdal
while loading C:\Users\appveyor\.julia\v0.6\GDAL\deps\build.jl, in expression starting on line 33
```
Example of full AppVeyor log [here](https://ci.appveyor.com/project/visr/gdal-jl/build/1.0.63/job/amchjtlw8i5pk7gg#L84).

Don't know what is causing this, and don't have a 32 bit Windows at hand to test. Don't know if Linux 32 bit builds are working, or if this is only for Windows.

claireh93:
I also seem to be having this problem when I try to build GDAL on a 64 bit linux machine running Ubuntu 14.04:
```julia
Warning: 'conda-forge' already in 'channels' list, moving to the top
Fetching package metadata .............
Solving package specifications: .

# All requested packages already installed.
# packages in environment at /home/claireh/.julia/v0.6/Conda/deps/usr:
#
gdal                      2.2.3                    py27_0    conda-forge
========================================================[ ERROR: GDAL ]========================================================

LoadError: Provider BinDeps.PackageManager failed to satisfy dependency libgdal
while loading /home/claireh/.julia/v0.6/GDAL/deps/build.jl, in expression starting on line 36

===============================================================================================================================

=======================================================[ BUILD ERRORS ]========================================================

WARNING: GDAL had build errors.

 - packages with build errors remain installed in /home/claireh/.julia/v0.6
 - build the package(s) and all dependencies with `Pkg.build("GDAL")`
 - build a single package by running its `deps/build.jl` script

===============================================================================================================================
```
When I look in the Conda library, libgdal seems to be there, but Julia doesn't seem to find it? Any advice you could give would be much appreciated!

visr:
Hmm I'm not sure, I've seen the same issue on Travis, so something is wrong.
It seems indeed like the conda installation is fine but then it cannot find it. Does it work if you start julia as such:
`LD_LIBRARY_PATH="/home/claireh/.julia/v0.6/Conda/deps/usr/lib/" julia`
And then `Pkg.test("GDAL")`?

[...]

PR #45, opened 2018-03-27
Fix travis builds

evetion:
After building, thus setting up conda. Rerun Julia with the Conda installation as its dynamic library path. This circumvents #42


visr:
Great, thanks. At least this lets us run the tests while we work on a fix for the builds.

PR #47, opened 2018-06-11
handle Ptr{Cstring} better

### LINKED ISSUE #13: Process Ptr{Cstring} return type to Vector{UTF8String}
Some functions return a Ptr{Cstring} for a list of strings. This should be converted to a list of strings. The current behavior is to return a string, only the first in the list.

This came up in
https://github.com/visr/GDAL.jl/pull/12#issuecomment-195908558

The conversion can be done as such
https://github.com/yeesian/GDALUtils.jl/blob/aa799a9a50bf7323f3caa9c3e1468c48f951906a/src/metadata.jl#L48-L60

visr:
fixes #13 

Commit 5069a757953e4d07b71fae9c854f6d789a6753f9 is perhaps a bit more controversial, so it would be good if @yeesian could sign off on that change. Not sure why we were using `Ptr{Ptr{UInt8}}` instead of `Ptr{Cstring}`.

### DIFF
--- a/gen/rewriter.jl
+++ b/gen/rewriter.jl
@@ -163,8 +163,6 @@ function rewriter(ex::Expr)
         if intype in opaquepointers
             # in C the opaque pointers are all typedef void *, i.e. Ptr{Void}
             intypes[i] = :(Ptr{Void})
-        elseif intype == :(Ptr{Cstring})
-            intypes[i] = :StringList  # see misc.jl
         end
     end
     # modify ccall return type or wrap entire ccall based on return type
--- a/gen/rewriter.jl
+++ b/gen/rewriter.jl
@@ -38,6 +38,7 @@ const custom_rename = Dict{String, String}(
     # prevent overwriting Base functions
     "gdalinfo" => "Base.info",
     "ogr_g_isempty" => "Base.isempty",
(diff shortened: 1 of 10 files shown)


visr:
ArchGDAL passes tests with this diff:

```diff
diff --git a/src/ogr/geometry.jl b/src/ogr/geometry.jl
index 9e0a1cd..b6764f5 100644
--- a/src/ogr/geometry.jl
+++ b/src/ogr/geometry.jl
@@ -216,11 +216,11 @@ wkbsize(geom::AbstractGeometry) = GDAL.wkbsize(geom.ptr)

 "Convert a geometry into well known text format."
 function toWKT(geom::AbstractGeometry)
-    wkt_ptr = Ref{Ptr{UInt8}}()
+    wkt_ptr = Ref(Cstring(C_NULL))
     result = GDAL.exporttowkt(geom.ptr, wkt_ptr)
     @ogrerr result "OGRErr $result: failed to export geometry to WKT"
     wkt = unsafe_string(wkt_ptr[])
-    GDAL.C.OGRFree(Ptr{UInt8}(wkt_ptr[]))
+    GDAL.C.OGRFree(pointer(wkt_ptr[]))
     wkt
 end
```

yeesian:
LGTM

Issue #18, opened 2016-03-17
Driver Management


yeesian:
https://sgillies.net/2014/01/22/rethinking-driver-management-in-fiona-and-rasterio.html

Maybe we should not [register all the drivers](https://github.com/visr/GDAL.jl/blob/master/src/GDAL.jl#L47-L53) when importing the module, since we do not call `GDALDestroyDriverManager()`?

visr:
Sean writes about problems with Python's `atexit` module.
Do we also have these problems? Could we for instance add a finalizer to the module itself? I tried some things, but not sure how to test if it is working.

```
julia> module A
       f(A) = println("cleaning up...")
       finalizer(A, f)
       end
A

julia> module A; end
WARNING: replacing module A
A

julia> gc()
cleaning up...error in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
```

```julia
finalizer(GDAL, GDAL.destroydrivermanager)
finalizer(GDAL, GDAL.destroy)
```

yeesian:
I don't think we can: https://github.com/JuliaLang/julia/issues/8764

visr:
Ah thanks for the link. Ok I agree on the side-effect not being very clean. Shall we just remove the driver register functions from the `__init__`, or also provide a context manager function in this package?

yeesian:
I think removing the driver registration functions from `__init__` should be fine. There still isn't a particularly clean solution to (automatically) decide which functions to provide do-block equivalents for.

meggart:
I am trying to catch up a bit with your progress, so sorry for stupid questions. Would this mean that the driver registration and destruction would have to be handled by the user? How would a downstream package handle this?

yeesian:
Yup,

This would mean that driver registration and destruction will have to be handled by downstream packages/users, e.g.

```julia
import GDAL
GDAL.allregister()
...
GDAL.destroydrivermanager()
```

A downstream package might provide a do-block context e.g.
```module FriendlierGDAL
import GDAL
function registerdrivers(F::Function)
   GDAL.allregister()
   try
      f()
   finally
      GDAL.destroydrivermanager()
   end
end
```

Such that users can do something like
```julia
import FriendlierGDAL
with FriendlierGDAL.registerdrivers() do
   ...
end
```

PR #53, opened 2018-07-29
make this example work on its own

SimonDanisch:


visr:
Oh, sorry about that! I just added a note to the readme about that in 1b713963085deaf497ce5bf8013e010b1d04a7a6.

This function actually used to be in the `__init__` function, but we removed it after the discussion in #18. If you have other ideas about how to go about this, let us know, we can reconsider. The `FriendlierGDAL` exists already, it is `ArchGDAL`, though it still needs to be registered.

SimonDanisch:
Fair enough! ;) Thanks for the qick reply!

PR #57, opened 2018-09-19
README: a few more words on the GDALBuilder

visr:
Fixes https://github.com/JuliaGeo/GDALBuilder/issues/6#issuecomment-422791582

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This package is a binding to the C API of GDAL/OGR. It provides only a C style u
 Other packages can build on top of this to provide a more Julian user experience. See for example [ArchGDAL.jl](https://github.com/yeesian/ArchGDAL.jl).
 
 ## Installation
-This package is registered, so add it using `Pkg`. This will also download GDAL binaries created by [GDALBuilder](https://github.com/JuliaGeo/GDALBuilder). The number of available drivers/format is still limited but will grow over time, please make an issue at GDALBuilder if you need a change in the provided GDAL installation. The command line tools such as `gdalinfo` are also included, though note that these will need both `GDAL_DATA` to be set and `Sys.BINDIR` must be in your path.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,7 @@ This package is a binding to the C API of GDAL/OGR. It provides only a C style u
 Other packages can build on top of this to provide a more Julian user experience. See for example [ArchGDAL.jl](https://github.com/yeesian/ArchGDAL.jl).
 
 ## Installation
-This package is registered, so add it using `Pkg`. This will also download GDAL binaries created by [GDALBuilder](https://github.com/JuliaGeo/GDALBuilder).
+This package is registered, so add it using `Pkg`. This will also download GDAL binaries created by [GDALBuilder](https://github.com/JuliaGeo/GDALBuilder). The number of available drivers/format is still limited but will grow over time, please make an issue at GDALBuilder if you need a change in the provided GDAL installation. The command line tools such as `gdalinfo` are also included, though note that these will need both `GDAL_DATA` to be set and `Sys.BINDIR` must be in your path.
```
yeesian: It'll be nice if we can also mention what instructions to run to check for `GDAL_DATA` and `Sys.BINDIR`

visr: Yeah indeed. But then perhaps we should move it to the back and extend it bit. Not every user that just wants to install it needs to read that.

yeesian: Perhaps a troubleshooting section (e.g. in [IJulia's README](https://github.com/JuliaLang/IJulia.jl#troubleshooting))?

visr: Yeah something like that sounds good!

visr: I pushed a troubleshooting section, would appreciate if you could have another look.

In the next version of the build, more utilies wil be present in deps.jl due to https://github.com/JuliaGeo/GDALBuilder/commit/fc8f1a64627332876acf5ba4bf1b00c01d0fe965

yeesian:
LGTM

PR #60, opened 2018-10-16
[WIP] switch to a new build with GDAL 2.3.2 and add curl and sqlite3

visr:
I'd still like to add a test to verify that new formats are now available, but let's see if this passes tests.

This does not include regenerating the wrapper, probably best to do that separately.


visr:
Closing in favor of #66

PR #66, opened 2019-05-20
switch to a new build with GDAL 3.0.0

visr:
And add ~~curl~~ and sqlite3

I just noticed that despite

https://github.com/JuliaGeo/GDALBuilder/blob/5853e073b2c1e46987c905f2093037819da09c2c/build_tarballs.jl#L42-L43

The Windows tarballs here

https://github.com/JuliaGeo/GDALBuilder/releases/tag/v3.0.0-1

are statically linked, that is why the tarballs are ~1GB. So not quite there yet, we need a new build.


visr:
@evetion do you have any clues? Might this be because libtool is disabled on Windows?

Looking at the build log [deps/usr/logs/GDAL.log](https://gist.github.com/visr/85951a6e24f77a52a9680f33a2e56d87)

```
checking if libtool supports shared libraries... yes
checking whether to build shared libraries... yes
checking whether to build static libraries... no
checking whether the /opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-gcc linker (/opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-ld) supports shared libraries... yes
checking whether the /opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-g++ -std=c++11 linker (/opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-ld) supports shared libraries... yes
checking for /opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-g++ -std=c++11 -shared ... no(3)
checking for ld -shared ... no
```

visr:
Currently having issues with PROJ_LIB. In the last commit I tried both by environment variable and directly using `proj_context_set_search_paths`. Neither seems to be working (whilst setting a global PROJ_LIB env var outside julia does).

The PROJ calls do make PROJ find the right database at least inside the `__init__` function, that is why it shows `dbpath = "C:\\Users\\visser_mn\\.julia\\dev\\GDAL\\deps\\usr\\share\\proj\\proj.db"`. But then in the tests it is GDAL's `importfromepsg` leading to this PROJ error:

```
ERROR: LoadError: GDALError (CE_Failure, code 1):
        PROJ: proj_create_from_database: Cannot find proj.db

Stacktrace:
 [1] gdaljl_errorhandler(::GDAL.CPLErr, ::Int32, ::Cstring) at C:\Users\visser_mn\.julia\dev\GDAL\src\error.jl:26
 [2] importfromepsg(::Ptr{GDAL.OGRSpatialReferenceH}, ::Int64) at C:\Users\visser_mn\.julia\dev\GDAL\src\ogr_srs_api.jl:136
```

evetion:
Congratulations on this amount of work! We now have a working 3.0 version, including PROJ6. Now the work begins on more drivers...

Not sure about the errors on the x86 windows builds, but they seem unrelated to GDAL itself.

Issue #70, opened 2019-06-01
Wrap GDAL 3.0


visr:
Now that we use the GDAL 3.0 library, we should rerun the wrappers under gen/ and bring the code back in sync with the libraries. IMO it would be best to do that before tagging a new release, as long as it doesn't take too long.

For adding the docstrings some work is needed anyway because of the Clang.jl changes. I'd like to see if we can make this code simpler via https://discourse.julialang.org/t/ann-automaticdocstrings-jl/24675.

PR #72, opened 2019-08-06
rewrite and rerun wrap_gdal.jl

visr:
This only takes care of updating the C submodule to GDAL 3. The rest still needs to be done by updating and rerunning `gen/rewriter.jl`. But this can be done in a separate PR.

@yeesian for a quick check?

### DIFF
--- a/src/C/cpl_vsi.jl
+++ b/src/C/cpl_vsi.jl
@@ -1013,7 +1013,7 @@ function VSIGetDiskFreeSpace(pszDirname)
 end
 
 """
-    VSIInstallMemFileHandler(void) -> void
+    VSIInstallMemFileHandler() -> void
 
 Install "memory" file system handler.
 """
--- a/src/C/cpl_vsi.jl
+++ b/src/C/cpl_vsi.jl
@@ -1,58 +1,50 @@
-# Julia wrapper for header: /usr/local/include/cpl_vsi.h
-# Automatically generated using Clang.jl wrap_c, version 0.0.0
-
-
+# Julia wrapper for header: cpl_vsi.h
+# Automatically generated using Clang.jl
 
 """
     VSIFOpen(const char * pszFilename,
              const char * pszAccess) -> FILE *
 """
-function VSIFOpen(arg1, arg2)
-    ccall((:VSIFOpen, libgdal), Ptr{FILE}, (Cstring, Cstring), arg1, arg2)
+function VSIFOpen()
+    ccall((:VSIFOpen, libgdal), Ptr{Cint}, ())
 end
-
 
 """
     VSIFClose(FILE * fp) -> int
 """
 function VSIFClose(arg1)
-    ccall((:VSIFClose, libgdal), Cint, (Ptr{FILE},), arg1)
+    ccall((:VSIFClose, libgdal), Cint, (Ptr{Cint},), arg1)
 end
-
 
 """
     VSIFSeek(FILE * fp,
              long nOffset,
              int nWhence) -> int
 """
-function VSIFSeek(arg1, arg2::Clong, arg3::Cint)
-    ccall((:VSIFSeek, libgdal), Cint, (Ptr{FILE}, Clong, Cint), arg1, arg2, arg3)
+function VSIFSeek(arg1, arg2, arg3)
+    ccall((:VSIFSeek, libgdal), Cint, (Ptr{Cint}, Clong, Cint), arg1, arg2, arg3)
 end
-
 
 """
     VSIFTell(FILE * fp) -> long
 """
 function VSIFTell(arg1)
-    ccall((:VSIFTell, libgdal), Clong, (Ptr{FILE},), arg1)
+    ccall((:VSIFTell, libgdal), Clong, (Ptr{Cint},), arg1)
 end
-
 
 """
     VSIRewind(FILE * fp) -> void
 """
 function VSIRewind(arg1)
-    ccall((:VSIRewind, libgdal), Cvoid, (Ptr{FILE},), arg1)
+    ccall((:VSIRewind, libgdal), Cvoid, (Ptr{Cint},), arg1)
 end
-
 
 """
     VSIFFlush(FILE * fp) -> void
 """
 function VSIFFlush(arg1)
-    ccall((:VSIFFlush, libgdal), Cvoid, (Ptr{FILE},), arg1)
+    ccall((:VSIFFlush, libgdal), Cvoid, (Ptr{Cint},), arg1)
 end
-
 
 """
     VSIFRead(void * pBuffer,
@@ -64,7 +56,6 @@ function VSIFRead()
     ccall((:VSIFRead, libgdal), Cint, ())
 end
 
-
 """
     VSIFWrite(const void * pBuffer,
               size_t nSize,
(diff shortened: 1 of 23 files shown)


visr:
This takes care of one half of #70.

src/C/cpl_vsi.jl
```diff
@@ -1,59 +1,51 @@
...
+# Automatically generated using Clang.jl
 
 """
     VSIFOpen(const char * pszFilename,
              const char * pszAccess) -> FILE *
 """
-function VSIFOpen(arg1, arg2)
-    ccall((:VSIFOpen, libgdal), Ptr{FILE}, (Cstring, Cstring), arg1, arg2)
+function VSIFOpen()
+    ccall((:VSIFOpen, libgdal), Ptr{Cint}, ())
```
yeesian: what happened to the arguments here?

src/C/cpl_vsi.jl
```diff
@@ -75,60 +66,53 @@ function VSIFWrite()
...
-
 """
     VSIFGets(char * pszBuffer,
              int nBufferSize,
              FILE * fp) -> char *
 """
-function VSIFGets(arg1, arg2::Cint, arg3)
-    ccall((:VSIFGets, libgdal), Cstring, (Cstring, Cint, Ptr{FILE}), arg1, arg2, arg3)
+function VSIFGets(arg1, arg2, arg3)
+    ccall((:VSIFGets, libgdal), Cstring, (Cstring, Cint, Ptr{Cint}), arg1, arg2, arg3)
```
yeesian: Is it expected for `Ptr{FILE}` to be replaced by `Ptr{Cint}`?

src/C/cpl_vsi.jl
```diff
@@ -1135,28 +1339,25 @@ function VSITime(arg1)
...
 end
 
-
 """
     VSIGMTime(const time_t * pnTime,
               struct tm * poBrokenTime) -> struct tm *
 """
 function VSIGMTime(pnTime, poBrokenTime)
-    ccall((:VSIGMTime, libgdal), Ptr{Cvoid}, (Ptr{time_t}, Ptr{Cvoid}), pnTime, poBrokenTime)
+    ccall((:VSIGMTime, libgdal), Ptr{Ctm}, (Ptr{Cint}, Ptr{Ctm}), pnTime, poBrokenTime)
```
yeesian: What does `Ctm` stand for?

src/C/gdal.jl
```diff
@@ -345,53 +433,23 @@ Open a raster file as a GDALDataset.
...
-* **papszSiblingFiles**: NULL, or a NULL terminated list of strings that are filenames that are auxiliary to the main filename. If NULL is passed, a probing of the file system will be done.
-
-### Returns
-A GDALDatasetH handle or NULL on failure. For C++ applications this handle can be cast to a GDALDataset *.
-"""
-function GDALOpenEx(pszFilename, nOpenFlags::UInt32, papszAllowedDrivers, papszOpenOptions, papszSiblingFiles)
+function GDALOpenEx(pszFilename, nOpenFlags, papszAllowedDrivers, papszOpenOptions, papszSiblingFiles)
```
yeesian: did we lose the docstring here?

src/C/gdal.jl
```diff
@@ -766,118 +806,99 @@ end
...
-"""
-    GDALClose(GDALDatasetH hDS) -> friend void
-
-Close GDAL dataset.
-
-### Parameters
-* **hDS**: The dataset to close. May be cast from a "GDALDataset *".
-"""
-function GDALClose(arg1::GDALDatasetH)
+function GDALClose(arg1)
```
yeesian: did we lose the docstring here?

src/C/ogr_srs_api.jl
```diff
@@ -54,136 +60,96 @@ OGRSpatialReference destructor.
...
     ccall((:OSRImportFromEPSG, libgdal), OGRErr, (OGRSpatialReferenceH, Cint), arg1, arg2)
 end
 
-
 """
     OSRImportFromEPSGA(OGRSpatialReferenceH hSRS,
                        int nCode) -> OGRErr
 
-Initialize SRS based on EPSG GCS or PCS code.
+Initialize SRS based on EPSG CRS code.
```
yeesian: Maybe we should have a way of storing the results (function names and their docstrings), so that we can write tools in the future to generate a CHANGELOG from run to run. 😝

src/C/ogr_srs_api.jl
```diff
@@ -416,30 +376,55 @@ end
...
+### Parameters
+* **hSRS**: source SRS
+* **pszTargetProjection**: target projection.
+* **papszOptions**: lists of options. None supported currently.
+
+### Returns
+a new SRS, or NULL in case of error.
+"""
+function OSRConvertToOtherProjection(hSRS, pszTargetProjection, papszOptions)
+    ccall((:OSRConvertToOtherProjection, libgdal), OGRSpatialReferenceH, (OGRSpatialReferenceH, Cstring, Ptr{Cstring}), hSRS, pszTargetProjection, papszOptions)
```
yeesian: I wonder where we can find example code for new functions being introduced haha. Always takes me a long time to figure out a working example from scratch

src/C/cpl_vsi.jl
visr: Don't know what it stands for, but this is the definition. It's one of the ones that Clang.jl always writes:
const Ctm = Base.Libc.TmStruct

src/C/cpl_vsi.jl
visr: Don't know, it doesn't appear to have changed [in the docs](https://gdal.org/api/cpl.html#cpl__vsi_8h_1ae3cfa1605ce32e78fddb534bba7742f5). We should look into that, though that means digging into Clang.

src/C/cpl_vsi.jl
visr: I think so, yes, due to Clang.jl changes. `FILE` is no longer being wrapped and put into common.jl either.

src/C/gdal.jl
visr: Good catch, fixed in 5e8633e7a4ab6d834bb4c8cd356427a6f54c2113, I picked the wrong XML node with name GDALClose here.

src/C/gdal.jl
visr: Good catch, fixed in 5e8633e7a4ab6d834bb4c8cd356427a6f54c2113, I picked the wrong XML node with name GDALOpenEx here.

src/C/ogr_srs_api.jl
visr: Yeah this PR is unfortunately a conflation of a new GDAL and changes to the docstring generation.

It's quite easy to write all docstrings to JSON or whatever, was actually doing that earlier for this PR, but since it introduced an extra step I took it out.

yeesian: yeah we don't have to do that for this PR; was just thinking out loud

src/C/cpl_vsi.jl
visr: Wait the docs I linked to are VSIOpen**L**, not VSIOpen. So I guess that just changed in the API?

src/C/ogr_srs_api.jl
yeesian: not a blocker for this PR though.

visr: Yep I agree, don't know. This one is new since 2.3 apparently: https://gdal.org/api/ogr_srs_api.html?highlight=osrconverttootherprojection#_CPPv427OSRConvertToOtherProjection20OGRSpatialReferenceHPKcPPCKc

visr:
Thanks a lot, @yeesian! Good to merge?

yeesian:
yup!

src/C/cpl_vsi.jl
yeesian: linking here for now: https://lists.osgeo.org/pipermail/gdal-dev/2006-October/010478.html

PR #73, opened 2019-08-11
1.0 Release with latest GDAL3 bindings

### LINKED ISSUE #70: Wrap GDAL 3.0

visr:
1. stop removing prefixes from GDAL function names, thereby overloading them
2. removes the C submodule, everything is now together in the main module

Fixes #70.

I realize the function renaming is quite breaking. In the README.md I wrote
a bit about the changes and how to update code with the [renamer](https://gist.github.com/visr/0a2ad3fe92073345c93c2ca42f5f58a0#file-renamer-jl).

Before we had many methods of (for instance) destroy. We used multiple dispatch to select the right C function if you called `GDAL.destroy(x)`,
depending on the type of `x`:

```json
"destroy": [
        "gdaldestroy",
        "gdal_cg_destroy",
        "ogr_fld_destroy",
        "ogr_gfld_destroy",
        "ogr_fd_destroy",
        "ogr_f_destroy",
        "ogr_ds_destroy",
        "ogr_sm_destroy",
        "ogr_st_destroy",
        "ogr_stbl_destroy"
    ]
```

This only worked because we rewrote the wrapping code, to pretend that `Ptr{Cvoid}` was actually `Ptr{OGRFeatureH}`, where `OGRFeatureH` was an abstract type we put in [types.jl](https://github.com/JuliaGeo/GDAL.jl/blob/b75ac1c3a351ac934e35827289a0d522a1fa7005/src/types.jl). While it was nice to use dispatch here, it did not match well with the C API. Firstly, it meant that wrapping became a complicated puzzle which did not neccesarily work. In `types.jl` we used subtyping to simulate that some functions could take different types of inputs. Now it is up to the user to select the right function. The functions will accept everything that can be converted to the types that C expects. This is in line with the new default Clang.jl behavior. This more straightforward wrapping of the GDAL C API is a better fit for this package, allowing others such as ArchGDAL.jl to try to make it more Julian. As you can see the wrapping code is now much simpler as a result, which should make it easier to maintain this package.

Regarding the C submodule, since we now do less rewriting, there is no strong reason anymore to keep them separate, folding them together.


evetion:
I already like the title 👍

gen/doc.jl
```diff
@@ -72,7 +71,74 @@ function build_docstring(node::EzXML.Node)
     end
 end
 
+"Return the text of a subelement `el` of `node`"
 function text(node::EzXML.Node, el::AbstractString)
     s = findfirst(el, node)
     s === nothing ? "" : strip(nodecontent(s))
 end
+
+"Wrap the one or multiline docstring in appropriate quotes"
```
yeesian: nit: "Wrap the single- or multi-line docstring in appropriate quotes."

gen/wrap.jl
```diff
@@ -0,0 +1,187 @@
...
+# set up a global logger to log.txt to store the large amount of logging
+loghandle = open(joinpath(@__DIR__, "log.txt"), "w")
+logger = SimpleLogger(loghandle)
+global_logger(logger)
+
+# several functions for building docstrings
+include(joinpath(@__DIR__, "doc.jl"))
+
+# should be here if you pkg> build GDAL
+includedir = normpath(joinpath(@__DIR__, "..", "deps", "usr", "include"))
```
yeesian: can we throw an error with the comment "should be here if you pkg> build GDAL" as a message if this line fails?

gen/wrap.jl
```diff
@@ -0,0 +1,187 @@
...
+function rewriter(x::Expr)
+    if @capture(x,
+        function f_(fargs__)
+            ccall(fname_, rettype_, argtypes_, argvalues__)
+        end
+    )
+        # it is a function wrapper around a ccall
+
+        # lowercase the function name
+        f′ = Symbol(lowercase(String(f)))
```
yeesian: nit: can we do `f2` instead of `f'`?

gen/wrap.jl
```diff
@@ -0,0 +1,187 @@
...
+        # it works since we include common.jl, meaning we do need to run this twice
+        # in case a new run creates new const X = Cvoid
+        isptrvoid = try
+            eval(rettype) == Ptr{Cvoid}
+        catch  # UndefVarError
+            # it might still be true if common.jl is not yet complete, run twice to avoid this
+            false
+        end
+
+        cc′ = if rettype == :Cstring
```
yeesian: nit: can we do `cc2` instead of `cc'`? That way we don't run the potential interpretation of `cc'` as the unary operator `'` being applied to the variable `cc`.

gen/wrap.jl
```diff
@@ -0,0 +1,187 @@
...
+    end
+end
+
+# parse GDAL's Doxygen XML file
+const doc = readxml(xmlpath)
+
+# custom function that prevents overwriting GDAL.jl by mapping gdal.h to gdal_h.jl
+function header_outputfile(h)
+    stem = splitext(basename(h))[1]
+    stem = stem == "gdal" ? "gdal_h" : stem
```
yeesian: nit: I think it might be clearer to write it as
```julia
if splitext(basename(h))[1] == "gdal"
    return joinpath(output_dir, "gdal_h.jl")
else
    return joinpath(output_dir, "gdal.jl")
end
```

yeesian:
[`OGRDatumType`](https://github.com/JuliaGeo/GDAL.jl/blob/38ccd2d37664673b81b1a86e9947bd5fe9a2f6d2/src/common.jl#L577) seems to be missing now.

yeesian:
```julia
julia> GDAL.GDALPaletteInterp
Cenum GDAL.GDALPaletteInterp:
GPI_Gray = 0x00000000
GPI_RGB = 0x00000001
GPI_CMYK = 0x00000002
GPI_HLS = 0x00000003

julia> GDAL.gdalgetpaletteinterpretation(GDAL.GPI_CMYK)
ERROR: MethodError: no method matching unsafe_convert(::Type{Ptr{Nothing}}, ::GDAL.GDALPaletteInterp)
Closest candidates are:
  unsafe_convert(::Type{Ptr{Nothing}}, ::Base.RefValue{T}) where T at refvalue.jl:30
  unsafe_convert(::Type{Ptr{Nothing}}, ::Base.RefArray{T,A,R} where R where A<:(AbstractArray{T,N} where N)) where T at refpointer.jl:78
  unsafe_convert(::Type{Ptr{Nothing}}, ::Base.CFunction) at c.jl:36
  ...
Stacktrace:
 [1] gdalgetpaletteinterpretation(::GDAL.GDALPaletteInterp) at /Users/yeesian/.julia/dev/GDAL/src/gdal_h.jl:2723
 [2] top-level scope at none:0
```

evetion:
Do we have an overview of all (registered?) packages that have GDAL as a dependency? This release will break all those packages.

I'm actually not in favor of deprecating all individual functions, but we should have a general deprecation warning at package load.

visr:
> OGRDatumType seems to be missing now.

I cannot seem to find a trace of it in the GDAL 3 repository, so I think it really dissappeared.

gen/doc.jl
visr: Thanks. Also switched to `repr` for single lines, since for PROJ at least I had an issue with unescaped quotes.

gen/wrap.jl
visr: Yeah don't want it to be confusing. It was a prime though, not a quote.

gen/wrap.jl
visr: Yes that's not quite correct but I get what you're saying ;)

visr:
> ERROR: MethodError: no method matching unsafe_convert(::Type{Ptr{Nothing}}, ::GDAL.GDALPaletteInterp)

Good catch thanks. I don't quite understand it yet. Has ccall behavior changed? Maybe this method should be defined by the `@cenum` macro. I can ask over there.

yeesian:
> > ERROR: MethodError: no method matching unsafe_convert(::Type{Ptr{Nothing}}, ::GDAL.GDALPaletteInterp)
> 
> Good catch thanks. I don't quite understand it yet. Has ccall behavior changed? Maybe this method should be defined by the `@cenum` macro. I can ask over there.

That was a mistake on my end which led to a false alarm: I was calling the wrong GDAL method.

visr:
> Do we have an overview of all (registered?) packages that have GDAL as a dependency? This release will break all those packages.

If https://discourse.julialang.org/t/reverse-dependencies-of-a-package/9774/5 is correct, it's just GeoArrays and ArchGDAL. Though of course there are plenty scripts out there that use this package directly.


> I'm actually not in favor of deprecating all individual functions, but we should have a general deprecation warning at package load.

I thought about deprecations for functions. But doing it properly (e.g. redirecting automatically to a working alternative) will be a difficult and large task. If anybody would be up for doing that it would be great, but I don't want to spend my time on it frankly. The general deprecation warning at package load, I'm not sure how we can do that such that it won't be annoying to users that have updated their code. The new version indicates breaking changes, and what they are and how to update can be found on the README.

yeesian:
I agree that (i) we should start having a better habit of maintaining a CHANGELOG as well as a smoother deprecation process, but also agree that (ii) it would have been too much work for us to provide that kind of user support in pre-1.0 versions.

We should have a version before 1.0 that provides deprecation warnings to get some experience in learning how the process works, but I don't realistically think it's going to happen for this PR, after having gone through https://github.com/yeesian/ArchGDAL.jl/pull/85 to get a sense for the complexity of the task.

yeesian:
I want to add that I like that the docstrings still refer to the original GDAL function names. It makes it easier for people who are doing a search based on the original names.

README.md
```diff
@@ -22,29 +22,31 @@ Docstrings are automatically inserted from the GDAL documentation. Note that the
...
 
-## Low-level and mid-level API
-Next to the functions available in the `GDAL` module, there is also a submodule `GDAL.C` available. `GDAL.C` is the low-level wrapping, without any error checking, function renaming or type wrapping. If you know what you are doing and don't want this package to get in your way, use this. Currently the `GDAL.C` module provides a more complete wrapping, though the intention is to make the complete user facing C API available in both modules.
+## Changelog
+
+Since GDAL.jl v0.2.0 the package has changed considerably. The GDAL C function names are no longer being shortened. This brought many complications forcing us to simulate GDAL's types to disambiguate methods. As an example, the old `GDAL.open` now has to be made explicit: use `GDAL.gdalopen`, `ogr_dr_open`, or `ogropen`. Other options are simpler, `GDAL.allregister` can only become `GDAL.gdalallregister`. To help with updating old code in a semi automatic fashion, look at [this script] which uses [MacroTools.jl](https://mikeinnes.github.io/MacroTools.jl/stable/sourcewalk/) to do function renaming.
```
yeesian: should be: `[this script](https://gist.github.com/visr/0a2ad3fe92073345c93c2ca42f5f58a0#file-renamer-jl)`

visr: Good catch, thanks!

visr:
> after having gone through [yeesian/ArchGDAL.jl#85](https://github.com/yeesian/ArchGDAL.jl/pull/85) to get a sense for the complexity of the task

Yeah that was quite the effort, nicely done! Probably that was the single largest updating effort haha.

visr:
Any objections to me merging this? How shall we do it with tagging releases? Do a simultaneous release of GDAL, ArchGDAL and GeoArrays?

evetion:
Let's do a simultaneous release. I've added a compat ```GDAL = ">=1.0"``` to my Project.toml.

meggart:
No objection from my side. I have a few projects depending on GDAL, with test the new version soon and will freeze GDAL version until then.

yeesian:
> Any objections to me merging this? How shall we do it with tagging releases? Do a simultaneous release of GDAL, ArchGDAL and GeoArrays?

Yeah, I think this package is ready for v1.0 :)

I think a new release of GDAL.jl should be made first -- modulo tests from downstream packages working locally. And we'll wait for green from Travis and AppVeyor before updating downstream packages ([ArchGDAL](https://github.com/yeesian/ArchGDAL.jl/pull/76) then [GeoArrays](https://github.com/evetion/GeoArrays.jl/pull/11); in the order of their dependencies).

What do you all think?

PR #74, opened 2019-08-22
reset the error handler before throwing a GDALError

### LINKED ISSUE #68: error handler not properly reset
Below example shows that `getdriverbyname` for non supported drivers returns null and doesn't throw. So far so good, this is apparently GDAL behavior since it is wrapped in `failsafe`. Calling `getdrivershortname` on this non existent driver does throw an error, fine. But then when I go back to `getdriverbyname` it does throw. Seems like while handling the GDAL error in error.jl, we don't properly reset the error state.

```julia
julia> using GDAL

julia> GDAL.allregister()

julia> driver = GDAL.getdriverbyname("ECW")
Ptr{GDAL.GDALDriverH} @0x0000000000000000

julia> driver = GDAL.getdriverbyname("ECW")
Ptr{GDAL.GDALDriverH} @0x0000000000000000

julia> driver = GDAL.getdriverbyname("NetCDF")
Ptr{GDAL.GDALDriverH} @0x0000000000000000

julia> GDAL.getdrivershortname(driver)
ERROR: GDALError (CE_Failure, code 10):
        Pointer 'hDriver' is NULL in 'GDALGetDriverShortName'.


Stacktrace:
 [1] gdaljl_errorhandler(::GDAL.CPLErr, ::Int32, ::Cstring) at C:\Users\visser_mn\.julia\dev\GDAL\src\error.jl:26
 [2] getdrivershortname(::Ptr{GDAL.GDALDriverH}) at C:\Users\visser_mn\.julia\dev\GDAL\src\gdal_h.jl:538
 [3] top-level scope at none:0

julia> driver = GDAL.getdriverbyname("NetCDF")
ERROR: GDALError (CE_Failure, code 10):
        Pointer 'hDriver' is NULL in 'GDALGetDriverShortName'.


Stacktrace:
 [1] failsafe(::Ptr{GDAL.GDALDriverH}) at C:\Users\visser_mn\.julia\dev\GDAL\src\error.jl:34
 [2] getdriverbyname(::String) at C:\Users\visser_mn\.julia\dev\GDAL\src\gdal_h.jl:398
 [3] top-level scope at none:0
```

visr:
Fixes #68. This means we cannot use `cplgetlasterror`* functions after an error anymore, but we can just get this information from the GDALError struct, as shown in the tests.

It seems quite neccesary to me to do this, since otherwise you get old error reports back, any time you call a new function that calls `failsafe` to check for errors, as shown in the issue.

Also expands and concentrates the error testing code.

### DIFF
--- a/src/error.jl
+++ b/src/error.jl
@@ -3,13 +3,17 @@ struct GDALError <: Exception
     class::CPLErr
     code::Cint
     msg::String
+    # reset GDAL's error stack on construction
+    function GDALError(class, code, msg)
+        cplerrorreset()
+        new(class, code, msg)
+    end
 end
 
 function GDALError()
     class = cplgetlasterrortype()
     code = cplgetlasterrorno()
     msg = cplgetlasterrormsg()
-    cplerrorreset()
     GDALError(class, code, msg)
 end
 
@@ -24,9 +28,7 @@ function gdaljl_errorhandler(class::CPLErr, errno::Cint, errmsg::Cstring)
     # function signature needs to match the one in __init__, and the signature
     # of the callback for a custom error handler in the GDAL docs
     if class in throw_class
-        msg = unsafe_string(errmsg)
-        cplerrorreset()
-        throw(GDALError(class, errno, msg))
+        throw(GDALError(class, errno, unsafe_string(errmsg)))
     end
     return C_NULL
 end
--- a/src/error.jl
+++ b/src/error.jl
@@ -9,6 +9,7 @@ function GDALError()
     class = cplgetlasterrortype()
     code = cplgetlasterrorno()
     msg = cplgetlasterrormsg()
+    cplerrorreset()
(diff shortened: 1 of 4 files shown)


src/error.jl
```diff
@@ -23,7 +24,9 @@ function gdaljl_errorhandler(class::CPLErr, errno::Cint, errmsg::Cstring)
     # function signature needs to match the one in __init__, and the signature
     # of the callback for a custom error handler in the GDAL docs
     if class in throw_class
-        throw(GDALError(class, errno, unsafe_string(errmsg)))
+        msg = unsafe_string(errmsg)
+        cplerrorreset()
```
evetion: You call `cplerrorreset` twice now, one here, one in the `GDALError`.

visr: That is a different path though. That is in the 0 arg constructor, which this one doesn't use.

Issue #19, opened 2016-03-25
Relax null checking for certain methods


yeesian:
I'm not sure if null checking should be performed (forcing downstream packages/users to wrap everything in a try-catch block) for cases in which you are expecting a null pointer to be returned. Case in point: https://github.com/visr/GDAL.jl/blob/master/src/ogr_api.jl#L3997-L4010

This came up when I was writing iterators for OGR feature layers, e.g. https://github.com/yeesian/GDALUtils.jl/blob/master/test/data1a.jl#L7-L13. I ended up using the C API instead: https://github.com/yeesian/GDALUtils.jl/blob/master/src/iteration.jl#L7-L8.

cc @visr @meggart

visr:
Perhaps the "certain methods" is best decided by GDAL itself? If we remove `checknull`, we will still get a `GDALError`, but only if GDAL calls [`CPLError`](http://www.gdal.org/cpl__error_8h.html#afda4d86428c1c533449ae6a69cdf430d) (right now even if it's only to report a `CE_Warning`).

I checked a few examples after setting `checknull(x) = x`.
Calling `GDALOpen` on a non existent file calls `CPLError` and thus throws. However `GDAL.getdriverbyname` for an unknown name just returns a null pointer, and so does your example, `GDAL.getnextfeature` if there is no next feature.

These examples seem reasonable, do you think we should just remove `checknull`?

yeesian:
I think (i) not checking for null pointers seems more like the exception rather than the rule, and (ii) making it the responsibility of users of (this package) to experience (and report) them seems like a good way of eliciting the range of edgecases.

I don't have a good understanding of how it works with CPLErrors yet. I propose we come up with a list of methods for which

1. **[Omission List]** we will not perform null checking. *I suspect this will tie in with notions of object ownership between the caller and callee, but we'll have a better sense once we have a few examples going*
2. **[OGRError List]** we will perform OGR error checks (e.g. https://github.com/yeesian/GDALUtils.jl/search?utf8=%E2%9C%93&q=result+%21%3D+GDAL.OGRERR_NONE)

And update them as we go? I'll go ahead and prepare the lists based on what I've seen in [GDALUtils](https://github.com/yeesian/GDALUtils.jl/) if we have a consensus here. I'm afraid I've reached the end of my spring break and won't have time to do a more systematic/comprehensive sweep of all the methods for now.

Waiting on responses from @visr @meggart, and whoever wishes to chime in!

[...]

PR #75, opened 2019-08-23
return nothing for NULL strings

### LINKED ISSUE #67: getattrvalue can be avoid throw expection and more julia way
`function getattrvalue(hSRS::Ref{OGRSpatialReferenceH}, pszName, iChild::Integer) `

return unsafe_string directly, if the result ccall returned is NULL, unsafe_string throw exception:

`ERROR: ArgumentError: cannot convert NULL to string`

call getattrvalue function need to handle exception,not easy to use.

I think getattrvalue function can add a judgment when returning, avoid to throw exception:
```
function getattrvalue(hSRS::Ref{OGRSpatialReferenceH}, pszName, iChild::Integer)
    result = ccall((:OSRGetAttrValue, libgdal), Cstring, (Ptr{Cvoid}, Cstring, Cint), hSRS, pszName, iChild)
    result != Cstring(Ptr{UInt8}(0)) ? unsafe_string(result) : nothing
end
```
I can call getattrvalue function by this way:
```
if getattrvalue(...) != nothing
...
else
...
end
```
or
`getattrvalue(...) != nothing && ...`

Instead of handling exceptions, do you think so ?

visr:
Fixes #67.

Currently this errors with `ArgumentError: cannot convert NULL to string` in `unsafe_string`, even though many functions are documented to return NULL in certain cases, that are not errors. This now works similarly to many Base functions like `findfirst`, that return `nothing` in case nothing is found.

### DIFF
--- a/gen/wrap.jl
+++ b/gen/wrap.jl
@@ -137,7 +137,7 @@ function rewriter(x::Expr)
         end
 
         cc2 = if rettype == :Cstring
-            :(unsafe_string($cc))
+            :(string_or_nothing($cc))
         elseif rettype == :(Ptr{Cstring})
             :(unsafe_loadstringlist($cc))
         elseif isptrvoid
--- a/gen/wrap.jl
+++ b/gen/wrap.jl
@@ -143,7 +143,9 @@ function rewriter(x::Expr)
         end
 
         cc2 = if occursin(r"^cpl.*error", String(f2))
-            # since we call these in aftercare, no aftercare to prevent calling itself
+            # We will take care of GDAL error states and memory freeing by
+            # calling `aftercare`, except for `cpl*error*` functions, as these
+            # will be called by `aftercare` itself, so we prevent a loop.
             if rettype == :Cstring
                 :(unsafe_string($cc))
             else
--- a/gen/wrap.jl
+++ b/gen/wrap.jl
@@ -78,6 +78,12 @@ const replace_exprs = Dict{Expr,Expr}(
     :(const VSIStatBufL = __stat64) => :(const VSIStatBufL = Cvoid),
 )
 
+# functions that return a Cstring, of which we know we must free them
+# because for instance the documentation says: "Free with CPLFree() or VSIFree()"
+# use the renamed (lowercased) function names here
+# if it is not in this list, it will not be freed, add function names to make it more complete
+const string_free_me = Symbol[:gdalinfo]
+
 """
 Custom rewriter for Clang.jl's C wrapper
 
@@ -136,14 +142,19 @@ function rewriter(x::Expr)
             false
         end
 
-        cc2 = if rettype == :Cstring
-            :(string_or_nothing($cc))
-        elseif rettype == :(Ptr{Cstring})
-            :(unsafe_loadstringlist($cc))
-        elseif isptrvoid
-            :(failsafe($cc))
+        cc2 = if occursin(r"^cpl.*error", String(f2))
+            # since we call these in aftercare, no aftercare to prevent calling itself
+            if rettype == :Cstring
+                :(unsafe_string($cc))
+            else
+                cc
+            end
+        elseif rettype == :Cstring
+            # boolean to let aftercare know if it should call vsifree
+            free = f2 in string_free_me
+            :(aftercare($cc, $free))
         else
-            cc
+            :(aftercare($cc))
         end
 
         # stitch the modified function expression back together
--- a/src/error.jl
+++ b/src/error.jl
@@ -82,5 +82,8 @@ function aftercare(ptr::Ptr{Cstring})
         i += 1
         cstring = unsafe_load(ptr, i)
     end
+    # TODO it seems that, like aftercare(::Cstring), we need to
+    # free the memory ourselves with CSLDestroy (not currently wrapped)
+    # not sure if that is true for some or all functions
     strings
 end
--- a/src/error.jl
+++ b/src/error.jl
@@ -22,6 +22,13 @@ function Base.showerror(io::IO, err::GDALError)
     println(io, err)
 end
 
+"""The custom GDAL error handling function that we set in __init__
+
+This function will get called by GDAL itself when it decides it wants to throw
+an error. We turn the error into a GDALError and throw it from Julia. This allows
+us to get GDALErrors that look similar to the ones that we throw when our own error
+checking code decides it should throw.
+"""
 function gdaljl_errorhandler(class::CPLErr, errno::Cint, errmsg::Cstring)
     # function signature needs to match the one in __init__, and the signature
     # of the callback for a custom error handler in the GDAL docs
@@ -43,6 +50,10 @@ end
 Handle anything returned from GDAL
 
 When values are returned from GDAL, always check if there was a failure using `maybe_throw`.
+If the failure was thrown by GDAL itself, it will not even get to `aftercare` and end up in
+`gdaljl_errorhandler`. However in many cases even though GDAL sets the error state to `CE_Failure`
+it will not throw the error. However we always do, to make sure not failure goes unnoticed.
+If an error might be expected, one can use `try .. catch` to handle this.
 
 Depending on the return type, we do extra work.
 """
--- a/src/error.jl
+++ b/src/error.jl
@@ -17,8 +17,6 @@ function GDALError()
     GDALError(class, code, msg)
 end
 
-const throw_class = (CE_Failure, CE_Fatal)
-
 function Base.showerror(io::IO, err::GDALError)
     err = string("GDALError (", err.class, ", code ", err.code, "):\n\t", err.msg)
     println(io, err)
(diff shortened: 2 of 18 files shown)


visr:
Thanks, your comments made me think about this again, and try something different. First my replies to your points:

> Is the name unsafe_string_or_nothing better?

From the `unsafe_string` docs: "This function is labeled "unsafe" because it will crash if p is not a valid memory address to data of the requested length." In this case I wouldn't call it unsafe since we know that GDAL allocated this memory.

> Maybe we could document this change somewhere?

Yeah, I think we just need to make a changelog for going from v0.2 to v1.0. I'd prefer at this point to do that all at once I think. (Going forward we should keep it up to date)

> You could also return an empty string for example, as that is at least the same type.

Yeah but then you cannot distinguish it from an actual empty string result. And since this method is used by Base as well it seems like a better choice.

> Does the C_NULL indicate different things for different function (i.e. an error or nothing)?

It seems like it, from the docs:

- CPLSerializeXMLTree: the document on success or NULL on failure.
- VSIGetFileSystemOptions: a string, which must not be freed, or NULL if no options is declared.


Especially the last point made me think, currently we only check for errors when we get a `Ptr{Cvoid}` back from GDAL that is a NULL pointer. Digging back a little, to Even Rouault's comment in https://github.com/JuliaGeo/GDAL.jl/issues/19#issuecomment-238096271:

> CPLError() can be called whether the return code of the function is successful or not, but I'd suggest that you can check with CPLGetLastErrorXXXX() when the return code is an error (or if you check always, do not error out if the error type was just a warning for example)

So the new commit here effectively switches to always checking for errors, but only throwing on `CE_Failure`. I took out `CE_Fatal` from the error checking codes since this is not in need of checking anyway, if this happens, it also crashes julia, just try running `GDAL.cplemergencyerror("things are bad")` if you can lose your julia session.

This also introduces a list of functions that return Cstring and are known to need their memory freed. For now I only put in `gdalinfo`. But looking around at some of the docs, see below, I noticed how it's really a mix of yes/no on freeing. So I believe the (current) default of not freeing is fine, will leak some memory. As we populate this list (`string_free_me` in `wrap.jl`), we can make it a little less leaky along the way.

- CPLStrdup: pointer to a newly allocated copy of the string. Free with CPLFree() or VSIFree().
- CPLGetPath: Path in an internal string which must not be freed. The string may be destroyed by the next CPL filename handling call. The returned will generally not contain a trailing path separator.
- OGR_L_GetName: the layer name (must not been freed)

yeesian:
I ran into something similar with https://github.com/JuliaGeo/Proj4.jl/blob/882b48a9ea75198bc3325a35a93cac36e3f1ceec/test/proj6api.jl#L11-L13 and agree with the motive of this PR, but might need a few days to properly review it (if my input is needed).

visr:
Yeah the main reason I requested your review was not to check if I implemented my ideas correctly, Maarten could check that as well, but to make sure you are on board with this direction. If you'd like to go over it properly, that's also welcome of course, no need to rush this in.

Even if this doesn't nail it 100%, we can easily opt out of checking for errors or free more strings going forward without breaking code. To me it was a good sign that I didn't have to change any tests, just add one for the returning `nothing` for NULL strings part.

src/error.jl
```diff
@@ -27,16 +25,62 @@ end
...
         throw(GDALError(class, errno, unsafe_string(errmsg)))
     end
     return C_NULL
 end
 
-"Throw an error if a pointer is null and GDAL reports an error"
-function failsafe(ptr)
-    if ptr == C_NULL && cplgetlasterrortype() in throw_class
+"Check the last error type and throw a GDALError if it is a failure"
+function maybe_throw()
```
evetion: This function is oddly familiar to the above `gdaljl_errorhandler`. Is the above still used/needed?
The name seems too generic too, it could be something like `potential_errorhandler` or similar?

src/error.jl
```diff
@@ -27,16 +25,62 @@ end
...
+    end
+end
+
+"""
+For string list pointers, load them to Vector{String}
+
+That is it expects a `StringList`, in the sense of the CPL functions,
+as a null-terminated array of strings.
+"""
+function aftercare(ptr::Ptr{Cstring})
```
evetion: Do these pointers never need to be freed like the string pointer above?

src/error.jl
```diff
@@ -27,16 +25,62 @@ end
...
+end
+
+"""
+Handle anything returned from GDAL
+
+When values are returned from GDAL, always check if there was a failure using `maybe_throw`.
+
+Depending on the return type, we do extra work.
+"""
+function aftercare(x)
```
evetion: The naming can be more descriptive here. `gdal_errorcheck_and_free` is ugly, but more readable.

gen/wrap.jl
```diff
@@ -136,14 +142,19 @@ function rewriter(x::Expr)
...
         end
 
-        cc2 = if rettype == :Cstring
-            :(unsafe_string($cc))
-        elseif rettype == :(Ptr{Cstring})
-            :(unsafe_loadstringlist($cc))
-        elseif isptrvoid
-            :(failsafe($cc))
+        cc2 = if occursin(r"^cpl.*error", String(f2))
+            # since we call these in aftercare, no aftercare to prevent calling itself
```
evetion: This feels very meta. Can it be reworded?

src/error.jl
```diff
@@ -27,16 +25,62 @@ end
...
+end
+
+"For string pointers, load them to String, and free them if we should."
+function aftercare(ptr::Cstring, free::Bool)
+    maybe_throw()
+    if ptr == C_NULL
+        return nothing
+    else
+        s = unsafe_string(ptr)
+        free && vsifree(convert(Ptr{Cvoid}, ptr))
```
evetion: Freeing memory utils could also be placed in a `memory.jl` file? `errors.jl` is just not the place I would expect them.

evetion:
> So the new commit here effectively switches to always checking for errors, but only throwing on `CE_Failure`. I took out `CE_Fatal` from the error checking codes since this is not in need of checking anyway, if this happens, it also crashes julia, just try running `GDAL.cplemergencyerror("things are bad")` if you can lose your julia session.

This seems it's worth its own issue? From testing (on stable GDAL), it can handle one emergencyerror, but not two.

```julia
julia> using GDAL

julia> GDAL.cplemergencyerror("things are bad")
ERROR: GDALError (CE_Fatal, code 1):
	things are bad

Stacktrace:
 [1] gdaljl_errorhandler(::GDAL.CPLErr, ::Int32, ::Cstring) at /Users/me/.julia/packages/GDAL/VqKWA/src/error.jl:26
 [2] cplemergencyerror(::String) at /Users/me/.julia/packages/GDAL/VqKWA/src/cpl_error.jl:14
 [3] top-level scope at none:0

julia> GDAL.cplemergencyerror("things are bad")
FATAL: things are bad

signal (6): Abort trap: 6
in expression starting at no file:0
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 1954108 (Pool: 1953578; Big: 530); GC: 3
[1]    81882 abort      /Applications/Julia-1.1.app/Contents/Resources/julia/bin/julia
```

src/error.jl
visr: Hmm `maybe_throw` is not an error handler though. The `gdaljl_errorhandler` is still used. If GDAL itself throws an error, `gdaljl_errorhandler` will catch it and throw a `GDALError`. `maybe_throw` is instead used from our side, and the only thing it does is maybe throw a `GDALError` if there is a `CE_Failure`.

src/error.jl
visr: Yeah they do at least sometimes, see for instance:

`CPLGetConfigOptions`: a copy of the list, to be freed with `CSLDestroy()`

I didn't add it here yet since it is an improvement that we can add at any time, and because that function lives in `cpl_string.h` which we don't wrap. Which doesn't mean of course we cannot just directly ccall the function. Would be good to make an issue as a reminder though.

src/error.jl
visr: Yeah, though that still wouldn't describe everything it does. Then it would have to be `gdal_errorcheck_and_free_and_load_strings` ;)
And if we add a new functionality, we have to rename the function. That is why I went for the less descriptive but more general `aftercare`, since we basically do anything we need to returned values, including error checking. One benefit of this is that if we want to tweak it, we just modify the `aftercare` definitions, and don't have to run the wrapper again every time.

gen/wrap.jl
visr: Ha what would be better? Without this you basically get StackOverflow errors, since `cplgetlasterrortype` calls `aftercare` which calls `cplgetlasterrortype` and so forth.

src/error.jl
visr: True that `error.jl` is not only about errors anymore. But it's less that 100 lines and I'd like to keep the `aftercare`s together. Not sure of a better name for `error.jl` though, since it's mostly that.

visr:
> This seems it's worth its own issue? From testing (on stable GDAL), it can handle one emergencyerror, but not two.

Yeah I noticed the same thing. Though I believe it's not even supposed to be able to handle it once? See also the comment here: 
https://github.com/JuliaGeo/GDAL.jl/blob/147f62425ec08bca95091bebd4b11ca32f739a22/test/error.jl#L42-L54

Could be it's own issue, but I've never encountered a `CE_Fatal`, so it's mostly hypothetical at this point.

src/error.jl
evetion: Ok, we can create an issue of it. Maybe add it as a `# todo` in the code?

gen/wrap.jl
evetion: The comments was for the comment! It just needs a small introduction for aftercare.
```julia
# We will take care of gdal error states and memory freeing (for now) by calling `aftercare`, except for `*error*` functions, as these will be called by `aftercare` itself, so we prevent a loop.
```

src/error.jl
evetion: Ah, handling C libraries is not easy. Would be nice to have your comment (about who throws what, calling which function) in the code here.

src/error.jl
visr: Yep! See 259a27c8ce78e53a078dc624cf5c3bec9a5b07f6 and #77.

src/error.jl
visr: 9f9a754f86c0b92f511c78ac40690b0a29390744

PR #81, opened 2019-12-23
use new GDAL build with working CURL and GEOS

visr:
See https://github.com/JuliaGeo/GDALBuilder/pull/12 for more details.

With thanks to @evetion for fixing some long standing build issues.


evetion:
Finally, this took way longer than I hoped it would. Now let's get to yeesian/ArchGDAL.jl#76 and evetion/GeoArrays.jl#11!

PR #85, opened 2020-03-09
Change Link to the build script in README

felixcremer:
GDALBuilder is deprecated and this uses now the build scripts provided by Yggdrasil


visr:
Good catch, thanks!

PR #87, opened 2020-04-11
Update Travis to also test downstream ArchGDAL.

evetion:


visr:
Great, thanks, this will help discovering issue in the build or breaking changes. Travis on Windows has some issues, but it seems to be on the Travis side. Appveyor works fine.

I added caching of artifacts as suggested in https://discourse.julialang.org/t/recommendation-cache-julia-artifacts-in-ci-services/35484. That should make up for some of the CI time increase here.

PR #102, opened 2020-11-19
fix PROJ_jll to 7.0.1

visr:
It looks like the latest PROJ build, 7.2.0, is causing the tests of GDAL to fail. If so, then this should pass on CI. Not sure if we'd want to merge this as is though, since on previous GDAL.jl versions it would still allow newer PROJ builds.

If you run into issues like:
```
ERROR: LoadError: InitError: could not load library "C:\Users\appveyor\.julia\artifacts\5ef006d7efdd27e9e2777126967ac7e86e2a0a1d\bin\libgdal-26.dll"
The specified module could not be found.
```
You can pin PROJ_jll in your own environment as a temporary workaround.
```
pkg> add PROJ_jll
pkg> pin PROJ_jll@7.0.1
```

I should note that the new PROJ_jll does seem to work fine with Proj4.jl, as that passes tests: https://github.com/JuliaGeo/Proj4.jl

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,7 @@ PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 CEnum = "0.2, 0.3, 0.4"
 GDAL_jll = "3.0.4"
 MozillaCACerts_jll = ">= 2020"
-PROJ_jll = "7.0.1"
+PROJ_jll = "= 7.0.1"
 julia = "1.3"
 
 [extras]


visr:
This is the problem:
```
using DependencyWalker
using GDAL_jll
Library(GDAL_jll.libgdal_path)
✗ libproj_7_0.dll (NOT FOUND)
```

It is looking for `libproj_7_0.dll` whereas it only has `libproj_7_2.dll`. So we need to update the GDAL build in Yggdrasil to only work with a fixed minor version of PROJ.

visr:
closing, since this is fixed by the GDAL_jll v3.0.4+5, based on https://github.com/JuliaPackaging/Yggdrasil/pull/2138.

PR #117, opened 2021-05-26
no need to set the certificates on 1.6+

visr:
Should fix https://github.com/yeesian/ArchGDAL.jl/issues/200


visr:
This doesn't work on Linux for julia 1.6+, throwing a "CURL error: Cert verify failed: BADCERT_NOT_TRUSTED"

I also found this comment: https://github.com/JuliaLang/julia/issues/39289#issuecomment-770084847.

Though it is probably preferable to just use https://github.com/JuliaLang/NetworkOptions.jl, and remove the MozillaCACerts_jll dependency, for all julia versions. NetworkOptions also installs on older julia versions. Quickly reading the README, it seems we could call `ca_roots`, and set it with `cplsetconfigoption` if the result is not `nothing`.

PR #119, opened 2021-06-04
update README on using the utilities

### LINKED ISSUE #86: update README on using the utilities
This section should be updated now that since #82 we use Artifacts and no longer have a `deps.jl`:
https://github.com/JuliaGeo/GDAL.jl/tree/v1.1.4#using-the-gdal-and-ogr-utilities

edit: I removed the incorrect information for now, but it would be great to update it and put it back

visr:
Fixes #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,13 +43,40 @@ Further usage documentation is not yet available, but the files `test/tutorial_r
 
 The bulk of this package is generated automatically by the scripts under `gen/`. For developer documentation regarding this process see `gen/README.md`.
 
-## Changelog
+## Using the GDAL and OGR utilities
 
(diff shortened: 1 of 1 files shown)


yeesian (approved):
Looks great!

yeesian:
I think one of the examples you gave on Zulip was also super helpful for others:

```
GDAL.ogr2ogr_path() do ogr2ogr
    run(`$ogr2ogr -f "ESRI Shapefile" outputfile.dbf inputfile.csv`)
end
```

visr:
Ah yes, thanks, I will add that example as well.

Do you think it would be useful to copy some of this documentation to https://yeesian.com/ArchGDAL.jl/dev/reference/#API-Utilities as well? Some of this is also available through libgdal, though sometimes these are just more convenient.

yeesian:
> Do you think it would be useful to copy some of this documentation to https://yeesian.com/ArchGDAL.jl/dev/reference/#API-Utilities as well?

Yeah, I think so! I've been trying to think through end-to-end workflows more, and these are great

yeesian:
> I've been trying to think through end-to-end workflows more, and these are great

In particular, in combination with https://github.com/yeesian/ArchGDAL.jl/issues/201 to set up wizards for introducing people to the different utilities.

PR #121, opened 2021-08-10
update to new Clang.jl

visr:
Continues the work in #120.
I wasn't able yet to run the Clang.jl generator yet, so this just moves in the bindings from #120. Still needs some cleanup, but I'd like to get it working first.

Some constants that were present before seem to be missing, like
https://github.com/JuliaGeo/GDAL.jl/blob/d44dc737ec16a60d1edb81bb71e0cb5f0666f165/src/common.jl#L537

And before we had
https://github.com/JuliaGeo/GDAL.jl/blob/d44dc737ec16a60d1edb81bb71e0cb5f0666f165/src/common.jl#L1168
Which now became `mutable struct OGRGeomFieldDefnHS end`. Probably it looks like changing these settings may be able to help here.

```toml
# if true, opaque pointers in function arguments will be translated to `Ptr{Cvoid}`.
opaque_func_arg_as_PtrCvoid = false

# if true, opaque types are translated to `mutable struct` instead of `Cvoid`.
opaque_as_mutable_struct = true
```

Though it could also be due to some of the not-so-nice rewriting we did [here](https://github.com/JuliaGeo/GDAL.jl/blob/ac66e15fe6ed00822595697525a1f195069c0490/gen/wrap.jl#L136-L145).

cc @melonedo @Gnimuc


visr:
Great, the two issues mentioned above were solved by integrating #122 from @melonedo in this branch. Now both the GDAL.jl and ArchGDAL.jl tests are passing on this branch.

I also rebased it on top of master and added the `time_t` workaround suggested in https://github.com/JuliaGeo/GDAL.jl/pull/122#issuecomment-896010981.

Only when I run wrap.jl, the amount of docstrings seems to be much less than when @melonedo runs it.

For instance this docstring vanished:
https://github.com/JuliaGeo/GDAL.jl/blob/b54e657182c8c059c725bb4f657d6ef3ef186469/gen/GDAL.jl#L38-L53
https://github.com/JuliaGeo/GDAL.jl/blob/e93d3dfd9c497d7b177cc89cb22b4ab4324eaf91/src/GDAL.jl#L36-L39

@melonedo if you run wrap.jl on this branch, do the docstrings come back? Is the issue perhaps due to differences in the doxygen.xml we use? I use this one currently, which should be generated with the same GDAL version that we currently wrap, 3.2.

[doxygen.zip](https://github.com/JuliaGeo/GDAL.jl/files/6962275/doxygen.zip)

Both however sometimes seem to have too many empty lines, like here between the parameters:
https://github.com/JuliaGeo/GDAL.jl/blob/e93d3dfd9c497d7b177cc89cb22b4ab4324eaf91/src/GDAL.jl#L764-L779

melonedo:
You can run **add_doc.jl** to add the docstrings. Now it's harder to simply inject docstrings, so I created a script to do it. Of course, that loses some comments about function pointers.

Parameters now has an empty line is because sometimes parameters may contain more than one paragraphs, in that case an empty line is necessary. I think I can improve the docstring generation here to only add empty lines when necessary.

visr:
Ah of course, thanks. I missed that add_docs.jl was not run automatically. Now calling it from wrap.jl in e995f293cffbe80d46a1adcec206493115390452.

> Parameters now has an empty line is because sometimes parameters may contain more than one paragraphs, in that case an empty line is necessary. I think I can improve the docstring generation here to only add empty lines when necessary.

I understand. It's definitely not needed for merging this.

PR #122, opened 2021-08-10
Update generator scripts 2

melonedo:
This commit adds macros and makes it work on Windows.

### DIFF
--- a/gen/GDAL.jl
+++ b/gen/GDAL.jl
@@ -1,0 +1,22033 @@
+module GDAL
+using GDAL_jll
+
+export GDAL_jll
+
+using CEnum
+
+"""
+    Stat(const char * pszFilename,
+         VSIStatBufL * pStatBuf,
+         int nFlags) -> int
+"""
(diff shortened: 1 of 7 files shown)


melonedo:
I generated in both linux and Windows, and I turns out `time_t` is different on the two platforms(`Clong` vs `Clonglong`).
Without more tests, I think the best choice might be using multi-platform configuration then.

cc @Gnimuc

Gnimuc:
If `time_t` is the only one that is system-dependent, then using multi-platform is a bit overkill. I guess it's safe to define it as `const time_t = Int` in the prologue.jl.

Gnimuc:
You could also take this as an example: https://github.com/maleadt/LLVM.jl/blob/master/res/missing_defs.jl

melonedo:
Thanks! I will try it later.

visr:
Thanks for this! I cherry picked both commits into #121. The `time_t` workaround is also used there.

PR #124, opened 2021-10-03
`gdaldestroy` registered first at exit in GDAL.jl/gen/epilogue.jl `__init__()`

mathieu17g:
@visr as discussed in https://github.com/yeesian/ArchGDAL.jl/pull/245#discussion_r720805777 here is a PR proposition for comments.

I have kept commented the registration at exit of `cplseterrorhandler(C_NULL)` and of a trial wih `gdaldumpopendatasets` for easy revert for the former and comment for the latter

### DIFF
--- a/gen/epilogue.jl
+++ b/gen/epilogue.jl
@@ -5,10 +5,14 @@ const GDAL_DATA = Ref{String}()
 const PROJ_LIB = Ref{String}()
 
 function __init__()
+    # Cleanup at exit
+    atexit(gdaldestroy)
+    # atexit(() -> gdaldumpopendatasets("/dev/stderr")) # TODO find a way to pass a pointer to stderr
+    
     # register custom error handler
     funcptr = @cfunction(gdaljl_errorhandler, Ptr{Cvoid}, (CPLErr, Cint, Cstring))
     cplseterrorhandler(funcptr)
-    atexit(() -> cplseterrorhandler(C_NULL))
+    # atexit(() -> cplseterrorhandler(C_NULL))
 
     # get GDAL version number
     versionstring = gdalversioninfo("RELEASE_NAME")


visr:
Oh it is nice if indeed `gdaldestroy` will take care of everything. Here is the code and docstring for it: https://github.com/OSGeo/gdal/blob/3c518a6ca/gdal/gcore/gdaldllmain.cpp#L55-L100

For the dump open datasets function, I hope we can use it to test the issues that we were seeing, so we make sure that `gdaldestroy` does what we expect. It would be nice to have such a test in GDAL.jl and not only in ArchGDAL.jl.

To put the dump open datasets function in the init function, I'm wondering if it is a good idea. For proper applications you want to always close things, but I'm doubting if we should print this for interactive users that may just want to try out some things.

So `gdaldestroy` doesn't close the open datasets?

gen/epilogue.jl
```diff
@@ -5,10 +5,14 @@ const GDAL_DATA = Ref{String}()
 const PROJ_LIB = Ref{String}()
 
 function __init__()
+    # Cleanup at exit
+    atexit(gdaldestroy)
+    # atexit(() -> gdaldumpopendatasets("/dev/stderr")) # TODO find a way to pass a pointer to stderr
```
visr, line 10: Didn't check, but could this work? `pointer_from_objref(stderr)`

mathieu17g: I have tested a few combination with `stdout`, `stderr` and their `struct` members. Everything led to a Julia crash:
- `gdaldumpdatasets(pointer_from_objref(stderr))` partially worked one time but did not print the opened dataset list, just GDALDumpOpenDatasets return result (number of opened datasets: `return static_cast<int>(poAllDatasetMap->size())`)
- `gdaldumpdatasets(pointer_from_objref(stdout))` printed a list of opened dataset before crashing Julia

I propose to clean epilogue.jl and raise an issue on the proper usage of `gdaldumpopendatasets`

gen/epilogue.jl
```diff
@@ -5,10 +5,12 @@ const GDAL_DATA = Ref{String}()
 const PROJ_LIB = Ref{String}()
 
 function __init__()
+    # Cleanup at exit
+    atexit(gdaldestroy)
```
yeesian, line 9: Does this work if the user doesn't call `gdalallregister()`?

mathieu17g: I guess that if you want to do something with GDAL you need at least one driver to be loaded.
Do you know a use case (in Julia, C or C++) where one user load only one or few drivers ?
I could test it.

yeesian: > Do you know a use case (in Julia, C or C++) where one user load only one or few drivers ?

There might be cases, e.g. https://trac.osgeo.org/gdal/ticket/6543 where they're not doing any I/O, and just using the spatial projection functionality.

Outside of that, I seldom see people load drivers individually over calling `gdalallregister()`. For reference: in https://gdal.org/tutorials/raster_api_tut.html#opening-the-file, it's said that "If for some applications it is necessary to limit the set of drivers it may be helpful to review the code from [gdalallregister.cpp](https://github.com/OSGeo/gdal/blob/master/gdal/frmts/gdalallregister.cpp). Python automatically calls GDALAllRegister() when the gdal module is imported."

yeesian: > Does this work if the user doesn't call `gdalallregister()`?

I've tested it locally and it seems to work fine. Feel free to resolve this comment thread, thanks!

mathieu17g: I have browsed [QGIS/src/crssync/main.cpp](https://github.com/qgis/QGIS/blob/master/src/crssync/main.cpp) and from there:
- `QgsApplication::init` in [QGIS/src/core/qgsapplication.cpp](https://github.com/qgis/QGIS/blob/master/src/core/qgsapplication.cpp)
-  `QgsCoordinateReferenceSystem::syncDatabase()` [QGIS/src/core/proj/qgscoordinatereferencesystem.cpp](https://github.com/qgis/QGIS/blob/master/src/core/proj/qgscoordinatereferencesystem.cpp)

without being able to find any GDAL functions to reproduce it with GDAL.jl.
It seems more deeply embedded. Difficult to derive a test.

yeesian: I'm quite convinced from the bug reports in https://github.com/OSGeo/gdal/blob/fec15b146f8a750c23c5e765cac12ed5fc9c2b85/gdal/gcore/gdaldllmain.cpp#L92-L99 that `gdaldestroy()` is supposed to be callable without needing to call `gdalallregister()` first.

yeesian (approved):

yeesian:
> So `gdaldestroy` doesn't close the open datasets?

To my understanding (see links below), it does:
* https://trac.osgeo.org/gdal/ticket/3954
* https://github.com/OSGeo/gdal/blob/fec15b146f8a750c23c5e765cac12ed5fc9c2b85/gdal/gcore/gdaldllmain.cpp#L86
* https://github.com/OSGeo/gdal/blob/939321f91aefb8b78b07080212b77bdba31d44da/gdal/gcore/gdaldrivermanager.cpp#L153-L335

visr:
Ah good, glad it also closes open datasets. So regarding this PR, I'm happy to have it. Though there are two points I still want to discuss:

- Should we move `gdalallregister()` back into GDAL.jl? All other GDAL config and initialisation is here, so maybe this should too?
- Can we test the effect of `gdaldestroy` here in some way? Testing the effect of `atexit` is probably difficult. We could just run `gdaldestroy` ourselves, but then we effectively brick our session, right? If tests are not feasible then so be it.

yeesian:
> Should we move `gdalallregister()` back into GDAL.jl? All other GDAL config and initialisation is here, so maybe this should too?

Do we know which initializations are heavyweight and which ones are not? Maybe this should be tracked in a ticket for discussion. I'm personally okay with it, but I'm not a particularly sophisticated user of GDAL

mathieu17g:
> What kind of tests did you run to determine error or no error? Running the whole ArchGDAL test suite?

Yes, with the sequence
1. Modification of `__init__()` in ArchGDAL.jl/src/ArchGDAL.jl
2. Pkg.update()
3. Pkg.test("ArchGDAL")

yeesian:
> on first run GDAL.gdalallregister() finishes in 3ms for me

@visr yeah I've always experienced it to be fast for myself too. If it matters, my vote's for it to run by default :)

yeesian:
I wasn't able to reproduce the `IOError` when following the example in https://github.com/JuliaGeo/GDAL.jl/pull/124#issuecomment-938069725 in a M1 Macbook:

```julia
  | | |_| | | | (_| |  |  Version 1.6.0 (2021-03-24)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

shell> pwd
/Users/yeesian/.julia/dev/GDAL/test

julia> using GDAL

julia> utmsmall = joinpath(@__DIR__, "data/utmsmall.tif")
"/Users/yeesian/.julia/dev/GDAL/test/data/utmsmall.tif"

julia> GDAL.gdalallregister()

julia> dataset = GDAL.gdalopen(utmsmall, GDAL.GA_ReadOnly)
Ptr{Nothing} @0x00007fb7528ce390

julia> rm(utmsmall)
```

However, I consistently get the following error message:

```julia
julia> GDAL.gdaldestroy()
CPLGetTLSList(): pthread_setspecific() failed!
```

PR #136, opened 2022-06-02
GDAL 3.5.0

visr:
https://github.com/OSGeo/gdal/blob/v3.5.0/NEWS.md

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "add2ef01-049f-52c4-9ee2-e494f65e021a"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "Wrapper for GDAL - Geospatial Data Abstraction Library"
-version = "1.3.0"
+version = "1.4.0"
 
 [deps]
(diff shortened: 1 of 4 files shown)


visr:
Fixed after rebuilding GDAL with a higher libcurl compat.

visr:
From the release notes:

> Add read/write OGR Parquet (Apache Parquet) and 'Arrow' (Apache Arrow IPC File/Feather or stream) drivers. Only in CMake builds

I didn't switch the build over to CMakie yet, so they are not available. Also haven't checked what dependencies are needed for them to build.

joa-quim:
I built GDAL with cmake a couple weeks ago and for geoparquet there is the extra dependency of the ``arrow`` lib (it's documented in GDAL) which looks like a complicated beast with many possible building options. I just used the basic ones and everything built, but did not do any further tests.

evetion:
So the cmake build would also need to depend on something like https://github.com/JuliaPackaging/Yggdrasil/pull/918

PR #145, opened 2023-01-01
Change Ptr{Cvoid} to Any for pProgressData.

evetion:
For https://github.com/yeesian/ArchGDAL.jl/pull/353


visr (approved):
Great, thanks. Glad you found out about this trick. I added the URL to the julialang blog as a reference, and ran the generator code.

It looks like the docstrings in 3.6.1 changed a bit and some docstrings got worse. The previous time the generator ran was for 3.6.0 so it shows some noise. But your code did it's job to convert to Any.

PR #160, opened 2023-09-19
add ecosystem integration tests

rafaqz:
This PR adds a downstream integration script to CI that will check packages that use GDAL with every pull request.

There will be a `GROUP=GDAL` environment variable available and tested package should use it to limit tests to only those effected by GDAL to cut the time CI here will take. For e.g. ArchGDAL.jl that may be all tests, for Rasters.jl it will be 10%. etc.

Please comment now or PR later to add packages that should be tested.


visr:
Nice! Can you remove or edit IntegrationTest.yml from #154? No need to test ArchGDAL twice.

rafaqz:
Haha didn't know about that one. Yes this is just a more generic way to do that, I'll remove it.

rafaqz:
My GDAL was a few commits behind and I didn't have that file yet, didn't mean to step on toes here!

.github/workflows/Integrations.yml
```diff
@@ -5,50 +5,53 @@ on:
...
+          - {user: rafaqz, repo: Rasters.jl, group: GDAL}
 
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v1
         with:
           version: ${{ matrix.julia-version }}
           arch: x64
-      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-buildpkg@latest
```
visr, line 31:
```suggestion
      - uses: julia-actions/julia-buildpkg@v1
```

Let's specify the version number like the other actions. If there are newer versions dependabot will let us know and we can upgrade on our own terms, see for instance #159.

visr (approved):
One minor comment, otherwise this is good to go.

PR #161, opened 2023-09-22
test GDAL 3.7.2

visr:
This in turn allows GDAL 3.7.2, which requires PROJ_jll v901 (which is PROJ 9.3).

I want to see about downstream failures.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,7 @@ PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 CEnum = "0.2, 0.3, 0.4"
 GDAL_jll = "301.600"
 NetworkOptions = "1.2"
-PROJ_jll = "900"
+PROJ_jll = "900, 901"
 julia = "1.6"
 
 [extras]


visr:
Some breakage on ArchGDAL, and the Rasters downstream test is failing because Rasters doesn't have GDAL as a direct dependency and therefor no compat set? @rafaqz this should be possible right? Otherwise, I guess it's also not terrible to add GDAL as a direct dependency as well.

```
Rasters [a3a2b9e3-a471-40c9-b274-f788e487c689]: Test Failed at /home/runner/.julia/packages/Aqua/CKSSt/src/deps_compat.jl:16
  Expression: result ⊜ true
   Evaluated: ⟪result: 😭 FAILED: Rasters [a3a2b9e3-a471-40c9-b274-f788e487c689]
    `/home/runner/work/GDAL.jl/GDAL.jl/downstream/Project.toml` does not specify `compat` for:
    GDAL
```
https://github.com/JuliaGeo/GDAL.jl/actions/runs/6277253339/job/17048615674?pr=161#step:6:877

rafaqz:
Yeah I can add a direct dep, more correct than using `ArchGDAL.GDAL` anyway

visr:
Yeah it'll probably be a while before ArchGDAL is pure-julia and can drop the GDAL dependency.

visr:
I'll release this version, which allows both old and new GDAL/GEOS/PROJ versions, as a patch release. This to unblock some issues using the wrapper packages together, like https://github.com/JuliaGeo/LibGEOS.jl/issues/182#issuecomment-1701789450.

Then I will wrap the new functions from GDAL 3.7. Wrapping those functions means I need to drop support for GDAL 3.6 since it won't have those methods, which means it is also dropping support for PROJ < 9.3. This shall be tagged as a new minor release.

PR #180, opened 2024-10-09
Add LERC_jll as a dependency

tpoisot:
See #179 - this is not using the most recent LERC version


felixcremer:
The integration test failures seem to be the same as on master.

tpoisot:
Fair enough - is there an ETA?

PR #182, opened 2024-10-10
Update TagBot.yml

ViralBShah:


visr (approved):

visr:
Thanks! It's unclear why this is failing on Julia 1.11 on MacOS and Linux. As far as I can see the dependencies are just like https://github.com/JuliaGeo/GDAL.jl/pull/181 which passes. Luckily I see #184 is ok, so let's move on with releasing that.

ViralBShah:
Agree. The packages seem to be working fine for me now.

PR #188, opened 2024-11-15
Include Arrow C structs.

evetion:
By including `ogr_recordbatch.h`. Let's see how far we can get in GeoDataFrames building on that.


visr (approved):
Looks good to me, thanks!

PR #193, opened 2025-07-01
Update to use GDAL_jll built with GDAL 3.11

joa-quim:


visr:
I'll close this in favor of #194 which also wraps GDAL 3.11.

PR #212, opened 2026-09-06
Fix CA certificate fallback for GDAL HTTPS

### LINKED ISSUE #203: SSL certificate problem on MacOS Julia 1.12
Noticed this test failure on CI in https://github.com/JuliaGeo/GDAL.jl/pull/202.

```
vsicurl access: Error During Test at /Users/runner/work/GDAL.jl/GDAL.jl/test/drivers.jl:68
  Got exception outside of a @test
  GDALError (CE_Failure, code 11):
  	CURL error: SSL certificate problem: unable to get local issuer certificate
  
  Stacktrace:
    [1] maybe_throw
      @ ~/work/GDAL.jl/GDAL.jl/src/error.jl:42 [inlined]
    [2] aftercare
      @ ~/work/GDAL.jl/GDAL.jl/src/error.jl:59 [inlined]
    [3] gdalopen(pszFilename::String, eAccess::GDAL.GDALAccess)
      @ GDAL ~/work/GDAL.jl/GDAL.jl/src/libgdal.jl:7456
```

It works fine in Julia 1.10.

asinghvi17:
Fixes #203.

On macOS with Julia 1.12, `NetworkOptions.ca_roots()` returns `nothing` because Julia's own TLS stack can use the Keychain. GDAL's libcurl build still needs a filesystem path, so `/vsicurl/` HTTPS requests fail with `SSL certificate problem: unable to get local issuer certificate`.

Use `NetworkOptions.ca_roots_path()` to provide GDAL with a filesystem fallback. Existing GDAL/libcurl certificate settings remain authoritative, certificate directories are passed through `GDAL_HTTP_CAPATH` rather than as a CA bundle file, and Julia environment settings are copied into GDAL config when Windows' separate C runtime cannot see them.

Validation:

- Full test suite on macOS aarch64, Julia 1.12.7: 170 passed
- Existing `/vsicurl/` HTTPS test succeeds against the Sentinel S3 URL
- Added coverage for default fallback, CA directories, existing GDAL configuration, and environment configuration
- [Core CI](https://github.com/JuliaGeo/GDAL.jl/actions/runs/34009233863) passes on all seven Julia/OS jobs
- The ArchGDAL downstream integration passes
- At the time of this update, the GeoDataFrames and Rasters downstream integrations are still running

The GADM downstream integration currently fails an unrelated table-schema assertion (13 columns returned, 12 expected). This change only affects CA configuration during initialization and does not touch dataset or schema handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,11 @@ you set two things:
 Inside of julia (2) is always the case, and (1) happens on loading the `GDAL` module, in its
 `__init__` function.
 
+GDAL's HTTPS support uses the certificate roots selected by Julia's `NetworkOptions`.
+On systems where the platform roots are not available as a file, this uses Julia's bundled
+certificate roots. Set `GDAL_CURL_CA_BUNDLE`, `CURL_CA_BUNDLE`, `SSL_CERT_FILE`, or
+`GDAL_HTTP_CAPATH` before loading GDAL to use custom certificate roots.
+
 ## Missing driver to support a format
 
 If you get an error such as the one below:
--- a/src/GDAL.jl
+++ b/src/GDAL.jl
@@ -32,6 +32,13 @@ function _set_ca_roots!()
     any(option -> cplgetconfigoption(option, C_NULL) !== nothing,
         _GDAL_CA_CONFIG_OPTIONS) && return
 
+    # On Windows, GDAL's C runtime may not see changes made through Julia's ENV.
+    for option in _GDAL_CA_CONFIG_OPTIONS
+        haskey(ENV, option) || continue
+        cplsetconfigoption(option, ENV[option])
+        return
+    end
+
     ca_path = ca_roots_path()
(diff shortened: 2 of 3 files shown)


evetion:
Can't you just put the following somewhere in the CI? Replace by networkoptions if you want.
```
using MozillaCACerts_jll
ENV["CURL_CA_BUNDLE"] = MozillaCACerts_jll.cacert
```

Note that I fixed this upstream, so it shouldn't happen anymore on 1.13, so only Julia 1.12.2-1.12.17 on macOS is affected.

visr:
Thanks for the PR. But now that Julia 1.13 is out, it feels a bit late to still bother with this. Closing, let me know if you disagree.
Folks stuck on Julia 1.12 can use the lines from https://github.com/JuliaGeo/GDAL.jl/pull/212#issuecomment-5558062013.

PR #213, opened 2026-09-11
Generate and require GDAL 3.13

asinghvi17:
Regenerates the wrappers against GDAL_jll 305.1300.301 (GDAL 3.13.3), bumps the compat bound to `GDAL_jll = "305.1300"`, and bumps the package version to 1.13.0, following the same pattern as #202 (3.12) and #194 (3.11).

## Wrapper changes

- New functions: `GDALCloseEx`, `GDALClearMemoryCaches`, dataset inter-band covariance matrix functions, mdarray overviews, `GDALHilbertCode`, `OGR_G_GetInvalidityReason`, `OGR_G_ConcaveHullOfPolygons`, `OGR_L_GetAttributeFilter`, `GDALGetColorInterpretationList`.
- Upstream renamed `GDT_Byte` to `GDT_UInt8`. The enum member is now `GDT_UInt8`; `GDT_Byte` remains available as a `const` alias, so ArchGDAL's `convert` tables keep working unchanged.
- `RASTERIO_EXTRA_ARG_CURRENT_VERSION` is now 3.
- A few platform-dependent constants (`CPL_FRMT_GB_WITHOUT_PREFIX`, `SIZEOF_UNSIGNED_LONG`, `VSIStatBufL`) reflect a Linux generation host. They were last generated on Windows in #190, so this returns them to the values they had before 3.10.

`gen/doxygen.xml` was regenerated from the v3.13.3 source with doxygen 1.15. Note for whoever runs the generator next: doxygen ≥ 1.9 emits a `Doxyfile.xml` alongside the per-file XML, and `combine_gdal_doxygen_xml.py` does not exclude it, so it has to be moved aside before combining or it can end up as the merged root.

## Testing

- GDAL.jl test suite: 157/157 on Julia 1.12, Linux x86_64.
- ArchGDAL 0.10.12 test suite against this branch: all functional tests pass. The only failure is Aqua's persistent-tasks check, which builds a fresh environment that resolves the *registered* GDAL.jl 1.12 and GDAL_jll 304, so it cannot pass until this release exists.
- Also exercised against a locally built GDAL 3.13.3 from the Yggdrasil recipe: new wrappers bind correctly, GeoTIFF write/read, warp via PROJ, and GeoPackage write/read all work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GRfocB47kS3dTCihR1P16C

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "add2ef01-049f-52c4-9ee2-e494f65e021a"
 keywords = ["GDAL", "IO"]
 license = "MIT"
 desc = "Wrapper for GDAL - Geospatial Data Abstraction Library"
-version = "1.12.0"
+version = "1.13.0"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
@@ -14,7 +14,7 @@ PROJ_jll = "58948b4f-47e0-5654-a9ad-f609743f8632"
 [compat]
 Aqua = "0.8"
 CEnum = "0.2, 0.3, 0.4, 0.5"
-GDAL_jll = "304.1200"
+GDAL_jll = "305.1300"
 NetworkOptions = "1.2"
 PROJ_jll = "902"
 Test = "<0.0.1,1"
--- a/gen/Manifest.toml
+++ b/gen/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.12.3"
+julia_version = "1.12.7"
 manifest_format = "2.0"
 project_hash = "d3fb5d354c0dd41c6ba04619267a742a45345dca"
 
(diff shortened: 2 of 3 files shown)


asinghvi17:
CI note: the core test matrix passes on all platforms (Julia 1.10, 1, and pre; Linux, macOS, Windows). The three red downstream jobs are not caused by this PR:

- **Rasters.jl** and **GeoDataFrames.jl** fail in Aqua's `stale_deps` / `deps_compat` checks because the integration workflow `Pkg.develop`s GDAL into their environments. Both fail identically on `main` (run 28235039113, June 26).
- **GADM.jl** fails `length(Tables.columnnames(row)) == 12` with 13. I ran the same `GADM.get("IND", depth=1)` under GDAL.jl 1.12 / GDAL 3.12.3 and under this branch / GDAL 3.13.3: both return the same 13 columns (`fid`, `geom`, plus 11 attributes). The `fid` column comes from the current ArchGDAL Tables interface, so GADM's hard-coded count is stale regardless of GDAL version.

visr (approved):
Thanks!

Issue #146, opened 2023-02-01
GeoParquet driver broken


visr:
Since https://github.com/JuliaPackaging/Yggdrasil/pull/5848, GDAL_jll comes with (Geo)Arrow and (Geo)Parquet support. Loading GeoArrow files seems to work fine, but GeoParquet is only partially working.

Here is a minimal reproducer. The dataset opens fine, we can get the only layer and ask for it's name, but as soon as we try to get the first feature, it crashes Julia with an `EXCEPTION_ACCESS_VIOLATION` and no further details.

```julia
using GDAL

path = "/vsicurl/https://github.com/opengeospatial/geoparquet/raw/v1.0.0-beta.1/examples/example.parquet"
dataset = GDAL.gdalopenex(path, GDAL.GDAL_OF_VECTOR, C_NULL, C_NULL, C_NULL)
layer = GDAL.gdaldatasetgetlayer(dataset, 0)
GDAL.ogr_l_getname(layer)  # -> "example"
feature = GDAL.ogr_l_getnextfeature(layer)  # segfault
```

evetion:
Works here (macOS) with the latest release.

visr:
Hmm still segfaults on Windows on Julia 1.11.1, so perhaps this is a Windows only issue.
The Arrow_jll version used for GDAL_jll is quite outdated, we use v10 whilst v18 is out, see also https://github.com/JuliaGeo/GDAL.jl/issues/171.

visr:
Unfortunately upgrading to Arrow v18 in https://github.com/JuliaPackaging/Yggdrasil/pull/10018 didn't help, the above example still gives a segfault:

```
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7ffa7f1c4d18 -- _Unwind_Resume at /workspace/srcdir/gcc-13.2.0/libgcc\unwind-seh.c:352
in expression starting at REPL[6]:1
Allocations: 2909 (Pool: 2900; Big: 9); GC: 0
```

PR #214, opened 2026-09-12
Test GeoParquet

visr:
https://github.com/JuliaGeo/GDAL.jl/issues/146 is now finally fixed in GDAL.jl 1.13.0!


asinghvi17:
Nice!!

